PR #1, opened 2017-03-16
Add option to run a group of jobs

harterrt:
Hey @vitillo, can you take a look at this and share your thoughts? 

In particular, can you comment on the best way to integrate this with Airflow? This revision adds a command line option to run groups of jobs for `daily` or `weekly` jobs. I'm thinking we may want to run each job in a separate Airflow node instead, so we can take advantage of Airflow's retries and jobs will fail independently. Are we OK with having a lot of small and similar configs in the Airflow repository?


python_etl/example_job.py
```diff
@@ -0,0 +1,39 @@
+from datetime import date, timedelta
```
vitillo: I think it would be better if every ETL job has its own folder, i.e. package, as some jobs might require multiple modules.

python_etl/main.py
```diff
@@ -1,40 +1,17 @@
...
-    total = sum(map(lambda x: x[1], results.items()))
-    # Print the OS and client_id counts in descending order:
-    for pair in sorted(results.items(), key=lambda x: -x[1]):
-        print "OS: {:<10} Percent: {:0.2f}%"\
-            .format(pair[0], float(pair[1])/total*100)
+job_periodicity_config = {
+    'daily': [
+        example_job.etl_job,
+    ],
+    'weekly': [],
```
vitillo: I think specifying the periodicity is best left to Airflow and ATMO.

scheduling/airflow.py
```diff
@@ -1,10 +1,18 @@
+import argparse
+
 from pyspark import SparkConf, SparkContext
 from pyspark.sql import SQLContext
 from python_etl import main
 
 if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='')
```
vitillo: I don't think we need this method at all. An ETL job could simply be invoked with e.g.:
```
python -m python_etl.example_job --start-date X --end-date Y
```

vitillo:
@harterrt I am leaning towards having individual tasks in Airflow.

harterrt:
Great, I think I agree. Closing this and moving to individual jobs.

PR #2, opened 2017-04-06
Use SparkSession instead of SparkContext

acmiyaguchi:
The default tests fail to run due to multiple spark contexts. This uses the consolidated SparkSession object from Spark 2.0 and creates a shared spark context.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,23 +1,29 @@
 import pytest
-from pyspark import SparkConf, SparkContext
+from pyspark.sql import SparkSession
 from python_etl import *
 
 # Initialize a spark context:
+
+
 @pytest.fixture(scope="session")
 def spark_context(request):
-    conf = SparkConf().setMaster("local")\
-        .setAppName("python_etl" + "_test")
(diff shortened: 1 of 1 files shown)


tests/test_main.py
```diff
@@ -1,24 +1,30 @@
...
+
+
 @pytest.fixture(scope="session")
 def spark_context(request):
-    conf = SparkConf().setMaster("local")\
-        .setAppName("python_etl" + "_test")
-    sc = SparkContext(conf=conf)
+    spark = (SparkSession
+             .builder
+             .appName("churn_test")
```
harterrt: Can you change this to "python_etl_test"?

acmiyaguchi: Done.

acmiyaguchi:
I think the issue occurs when there are tests running in parallel. I saw the issue when I added a new test file for a job that I'm porting over.

harterrt:
Sorry for the delay. There was a bug in my environment that was causing these tests to fail. This looks great, thanks for tackling this!

PR #5, opened 2017-04-12
Bug 1353215 - Port churn notebook to python_etl

acmiyaguchi:
[Bug 1353215 - Port churn notebook to python_etl](https://bugzilla.mozilla.org/show_bug.cgi?id=1353215)

This pull request adds churn to python_etl.

This will probably be launched from airflow like the telemetry-batch-view modules. I've chosen [click](http://click.pocoo.org/5/) to build the entry-point because of its goal of composable command line interfaces.

There are some more tests that need to be written before this can get merged in, but it would be nice to get some feedback in early.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/5)


harterrt:
Thanks, Anthony. This looks great! Travis and tox will make this much easier to work with!

I have worked with `click` before and it seems like a good fit here. It probably makes sense to address this in a future PR, but does it make sense to refactor the CLI into it's own module? It would be nice to share the spark context creation between jobs. 

I'll give this a more thorough look over the next couple days. Thanks again for improving this tool!

acmiyaguchi:
It would probably make sense to have a `@spark` decorator that we could share across module, kind of like the spark context in the test fixture. As for the cli, having an extra layer of indirection might be nice i.e. `python_etl churn <args>`.  Click supports [arbitrary nesting of commands](http://click.pocoo.org/5/quickstart/#nesting-commands), which would help keep jobs decoupled.

harterrt:
I took a closer look today, particularly with the upcoming work on the [engagement dashboard](https://docs.google.com/document/d/1QXyKFYcaoyOsbpeETSZxv4_03YPjgbcMlK2Br4wLnns/edit) in mind. Overall, this looks like a great fit and we should be able to adapt this format to that purpose. A few comments:

* It appears that we want metrics weighted by client_id, which sounds correct to me. We noted last week that unique client_id is difficult to track if we want to be able to aggregate the data further at analysis time. Does it make sense to track a HLL variable instead of a count? This adds a lot of complexity so I don't necessarily recommend doing this, but it would solve the following issues:
  * Some users could be in multiple (e.g.) `geo`'s. How do you handle this conflict currently? 
  * We can partition by submission date
* This dataset tracks E(X), E(X^2), and N so we can aggregate the data further and compute the variance at analysis time (E(X**2) - E(X)**2). This variance estimate can have [very large floating point errors](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance) for some distributions of data. This **does not appear to be a problem** for the measures we're tracking, but it's good to note for the record :). The error is approximately: `mean(x)**2/variance(x)*N*machine_error` so we should watch for metrics with big means and small variances.

Most of these comments can be addressed in future PRs, but I wanted to get them written out now while they're fresh in my head. Thanks for getting this code reviewed and tested, Anthony. This looks great!

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+        StructField('content',                 StringType(), True),
+        StructField('distribution_id',         StringType(), True),
+        StructField('default_search_engine',   StringType(), True),
+        StructField('locale',                  StringType(), True),
+        StructField('is_active',               StringType(), True),
+        StructField('n_profiles',              LongType(),   True),
+        StructField('usage_hours',             DoubleType(), True),
+        StructField('sum_squared_usage_hours', DoubleType(), True),
+        StructField('total_uri_count',         LongType(),   True),
+        StructField('unique_domains_count',    DoubleType(),   True)
```
mreid-moz: nit: please fix spacing in the last column

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+             .getOrCreate())
+
+    version = 'v2'
+    s3_path = "s3://{}/{}/{}".format(bucket, prefix, version)
+
+    # If this job is scheduled, we need the input date to lag a total of
+    # 10 days of slack for incoming data. Airflow subtracts 7 days to
+    # account for the weekly nature of this report.
+    week_start_date = start_date
+    if lag:
```
mreid-moz: Please factor out a function for computing lag and date-snapping and add some test coverage for different scenarios.

tests/test_churn.py
```diff
@@ -0,0 +1,389 @@
...
+    "locale":                "en-US",
+    "normalized_channel":    "release",
+    "profile_creation_date": 17181,
+    "submission_date_s3":    "20170115",
+    "subsession_length":     1000,
+    "subsession_start_date": "2017-01-15",
+    "sync_configured":       False,
+    "sync_count_desktop":    1,
+    "sync_count_mobile":     1,
+    "timestamp":             1491244610603260700,  # microseconds
```
mreid-moz: actual telemetry records have a `timestamp` in nanoseconds, not microseconds

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+            if datetime.strftime(client_date, "%Y%m%d") < week_start:
+                out["is_active"] = "no"
+        except ValueError:
+            pass
+    out["good"] = True
+    return out
+
+
+# Build the "effective version" cache:
+d2v = make_d2v(get_release_info())
```
mreid-moz: I don't think we should compute `d2v` map in the global namespace - at most it should be lazy-loaded when it's needed.

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+
+def fmt(d, date_format="%Y%m%d"):
+    return datetime.strftime(d, date_format)
+
+
+# ### Compute the aggregates
+#
+# Run the aggregation code, detecting files that are missing.
+#
+# The fields we want in the output are:
```
mreid-moz: Please move these field descriptions down next to the `churn_schema` declaration.

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+    week_start: datestring of this time period"""
+
+    week_start_date = datetime.strptime(week_start, "%Y%m%d")
+    week_end_date = week_start_date + timedelta(6)
+    week_start = fmt(week_start_date)
+    week_end = fmt(week_end_date)
+
+    # Verify that the start date is a Sunday
+    if week_start_date.weekday() != 6:
+        print("Week start date {} is not a Sunday".format(week_start))
```
mreid-moz: Should we raise an exception instead of printing a message here? I think we will want to treat a request to run on the wrong day as an ETL error.

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+        return
+
+    # If the data for this week can still be coming, don't try to compute the
+    # churn.
+    week_end_slop = fmt(week_end_date + timedelta(10))
+    today = fmt(datetime.utcnow())
+    if week_end_slop >= today:
+        print("Skipping week of {} to {} - Data is still arriving until {}."
```
mreid-moz: Similarly, should we upgrade this to an exception?

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+    # Apply some post-processing for other aggregates
+    # (i.e. unique_domains_count). This needs to be done when you want
+    # something other than just a simple sum
+    def average(total, n):
+        if not n:
+            return 0.0
+        return float(total) / n
+    average_udf = F.udf(average, DoubleType())
+
+    # Create new derived columns and any unecessary ones
```
mreid-moz: +drop :)

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+
+
+@click.command()
+@click.option('-d', '--date', 'start_date',
+              default=fmt(datetime.now() - timedelta(7)),
+              help="datestring in form YYYYmmdd")
+@click.option('-b', '--bucket',
+              default='net-mozaws-prod-us-west-2-pipeline-analysis',
+              help="output s3 bucket")
+@click.option('-p', '--prefix', default='telemetry-test-bucket/churn-test',
```
mreid-moz: I don't think the defaults for `--bucket` and `--prefix` make sense. If you want to use `s3://telemetry-test-bucket`, then that should go into the `--bucket` argument.

In fact, do we even want defaults for these? Maybe it would be better not to have defaults and simply bail out if these args are not present (date, bucket, prefix).

acmiyaguchi:
Thanks for the comments Ryan.

* A HLL variable would allow more granular analysis. I'm a bit unfamiliar with the specifics about doing aggregates as opposed to counts with HLL, but having the ability to create sets of clients and their intersections by day would allow this (or future) datasets to do more than [unbounded retention measureas](https://amplitude.com/blog/2016/08/11/3-ways-measure-user-retention/). For example, an N day retention requires recomputing for each N, whereas data indexed by HLL would just take the intersection of every Nth day.
* Multiple geo's per client are handled by taking the location of the latest client ping per week. This doesn't pose an issue in this case because information about specific clients don't carry over between weeks.
* Is variance an important measure for all aggregates? The only measure that takes into account sum and sum<sup>2</sup> is usage hours. URI counts and unique URI domains do not have any sort of variance measure included because they weren't requested. I assume this need is dictated by the analysis in the end.

python_etl/churn.py
acmiyaguchi: That sounds good, I'll probably make more use of exceptions now that this is its own module. I'm also probably going to switch over to using logging instead of print statements.

harterrt:
> Is variance an important measure for all aggregates?

For the engagement dashboard we'll probably need variance for all aggregates. The need is not explicitly listed, but the P1 goal is to understand whether a change is negative, which will require some type of hypothesis test.

tests/test_churn.py
acmiyaguchi: The `timestamp` field on main_summary seems to be in microsecond resolution. 

`spark.read.parquet('s3a://telemetry-parquet/main_summary/v3/submission_date_s3=20170401/sample_id=0').select('timestamp').show(truncate=False)`

python_etl/churn.py
```diff
@@ -0,0 +1,654 @@
...
+
+Tracked in Bug 1226379 [1]. The underlying dataset is generated via
+the telemetry-batch-view [2] code, and is generated once a day. The
+aggregated churn data is updated weekly.
+
+Due to the client reporting latency, we need to wait 10 days for the
+data to stabilize. If the date is passed into report through the
+environment, it is assumed that the date is at least a week greater
+than the report start date.  For example, if today is `20170323`, then
+this notebook `20170316` in the `environment.date`. This notebook will
```
acmiyaguchi: Done

python_etl/churn.py
acmiyaguchi: Moved closer to the convert function.

python_etl/churn.py
acmiyaguchi: Done

python_etl/churn.py
acmiyaguchi: Done

python_etl/churn.py
acmiyaguchi: Done

acmiyaguchi:
@mreid-moz RFAL

tests/test_churn.py
mreid-moz: It looks like nanos to me:
```
1491244610603260700
SSSSSSSSSSmmmuuunnn

S=Sec
m=millis
u=micros
n=nanos
```

tests/test_churn.py
```diff
@@ -0,0 +1,536 @@
...
+    StructField("profile_creation_date", LongType(),    True),
+    StructField("submission_date_s3",    StringType(),  False),
+    StructField("subsession_length",     LongType(),    True),
+    StructField("subsession_start_date", StringType(),  True),
+    StructField("sync_configured",       BooleanType(), True),
+    StructField("sync_count_desktop",    IntegerType(), True),
+    StructField("sync_count_mobile",     IntegerType(), True),
+    StructField("timestamp",             LongType(),    True),
+    StructField("total_uri_count",       IntegerType(), True),
+    StructField("unique_domains_count", IntegerType(), True)])
```
mreid-moz: nit: please align the columns here with the rows above

tests/test_churn.py
acmiyaguchi: You're right, I trusted this [epoch converter](https://www.epochconverter.com/) as little too much...

PR #6, opened 2017-04-12
Add .travis.yml and tox.ini for automated testing

acmiyaguchi:
This should let this repository be run under continuous integration using travis and codecov.io. I've tested this against my `churn-port` branch. 

This uses tox for setting up the environment and test-runner. As long as you have your `PYTHONPATH` set correctly with the pyspark and py4j dependencies, you should be able to run `tox` to run tests.

Alternatively, you can run pytest directly if you want more flexibility. The dependencies for pytest are in setup.py for setting up your own virtualenv.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language: python
+python:
+    - "2.7"
+cache:
+    directories:
+        - $HOME/spark
+addons:
+    apt:
+        packages:
+            - libsnappy-dev
+before_install: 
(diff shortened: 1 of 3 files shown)


harterrt:
Thanks, Anthony! Looks great!

Sorry for the slow review, I had some trouble getting this working on my local machine due to some bad system configs on my side. Nothing to do with your changes, but in case others have similar issues I will note that installing the dependencies using conda make things work much smoother.

PR #8, opened 2017-04-18
Add .tox/ and .coverage to gitignore

harterrt:


acmiyaguchi:
LGTM, both files should not get checked in.

PR #11, opened 2017-04-24
Add simple wrapper to pulse etl job

harterrt:
This allows us to schedule the job from Airflow. Later today I plan on
moving this code into the package and generalizing with a click CLI
wrapper.

### DIFF
--- a/scheduling/pulse.py
+++ b/scheduling/pulse.py
@@ -1,0 +1,15 @@
+import os
+from pyspark import SparkConf, SparkContext
+from pyspark.sql import SQLContext
+from python_etl.testpilot import pulse
+
+if __name__ == "__main__":
+    conf = SparkConf().setAppName('pulse_etl')
+    sc = SparkContext(conf=conf)
+    sqlContext = SQLContext(sc)
+
+    # Make bdist_egg available to executer nodes
+    sc.addPyFile(os.path.join('dist', os.listdir('dist')[0]))
+
+    # Run job
+    pulse.etl_job(sc, sqlContext)


scheduling/pulse.py
```diff
@@ -0,0 +1,15 @@
+import os
+from pyspark import SparkConf, SparkContext
+from pyspark.sql import SQLContext
+from python_etl.testpilot import pulse
+
+if __name__ == "__main__":
+    conf = SparkConf().setAppName('python_etl')
```
acmiyaguchi: This should probably be `pulse_etl`.

scheduling/pulse.py
```diff
@@ -0,0 +1,15 @@
...
+    sc = SparkContext(conf=conf)
+    sqlContext = SQLContext(sc)
+
+    # Make bdist_egg available to executer nodes
+    sc.addPyFile(os.path.join('dist', os.listdir('dist')[0]))
```
acmiyaguchi: Non-blocking for this review, but would we be able to distribute this library by default like `python_moztelemetry` in the airflow clusters to avoid specifying these files?

scheduling/pulse.py
```diff
@@ -0,0 +1,15 @@
...
+
+    # Run job
+    pulse.etl_job(sc, sqlContext)
```
acmiyaguchi: How does the date get calculated for this job? Many airflow jobs take a date specified by the scheduler via environment variable ([$date as instantiated by`ds_nodash`](https://github.com/mozilla/telemetry-airflow/blob/master/dags/main_summary.py#L24)). The scheduler is free to handle retries and backfill that don't necessarily match the current time, resulting in generally consistent results.

If you want to add this to the cli, Click has a neat feature to take an argument from the environment: `@click.argument('date', envvar='date')`.

harterrt: The date is always 'yesterday' for this job (calculated [here](https://github.com/mozilla/python_etl/blob/master/python_etl/testpilot/utils.py#L7)).

For context, I want to get this version of the job checked in to work around [ATMO Issue 326](https://github.com/mozilla/telemetry-analysis-service/issues/326). We've needed to run this job manually for that last couple of weeks and it's impeding analysis for our clients. 

I'm not too concerned with date consistency for now since this job runs in <1h. Retries should still be done on the same day. I plan on implementing a more fully featured CLI app this afternoon, but it may take some more time to test and implement. Thanks for the click tip. I'll be sure to include that!

scheduling/pulse.py
harterrt: We should be able to. I'll look into this today.

scheduling/pulse.py
harterrt: Done

sunahsuh:
Sorry I didn't get to this yesterday, but I was curious -- there's a [scheduling/airflow.py](https://github.com/mozilla/python_etl/blob/master/scheduling/airflow.py) file that seems to do 90% of this PR. Is there a particular reason we can't use it here?

harterrt:
@sunahsuh - airflow.py schedules [the example job](https://github.com/mozilla/python_etl/blob/master/python_etl/main.py). I wanted to keep this file as-is so we keep a bare-bones example to get new users started. 

I'd actually like to do away with the scheduling directory pretty soon. Ideally, the bash scripts would live in Airflow and the python scripts would sit in a little CLI inside the package.

PR #12, opened 2017-04-25
Some nitpicking to make the library API more Pythonic.

jezdez:


jezdez:
I've separated the code update from the renaming in two commits so it should be fair to review this. Let's see how happy Travis is..

scheduling/airflow.sh
```diff
@@ -4,7 +4,5 @@
 unset PYSPARK_DRIVER_PYTHON
 
 # Clone, install, and run
-git clone https://github.com/mozilla/python_etl.git python_etl
-cd python_etl
-pip install .
+pip install git+https://github.com/mozilla/python_mozetl.git
```
jezdez: that implies that the repo is renamed, which GitHub handles automatically with redirects both for the website as well as the Git repo

tests/conftest.py
```diff
@@ -18,6 +18,9 @@ def spark_context(request):
...
 
 
-def row_to_dict(row):
+@pytest.fixture(scope="session")
+def row_to_dict():
     """Convert pyspark.Row to dict for easier unordered comparison"""
-    return {key: row[key] for key in row.__fields__}
+    def func(row):
+        return {key: row[key] for key in row.__fields__}
+    return func
```
jezdez: This is a bit odd at first, but something that is common in pytest: a fixture that returns a function, so it can be shared by many tests and is centrally implemented.

tox.ini
```diff
@@ -7,13 +7,11 @@
 envlist = py27
 
 [pytest]
-addopts = --cov=python_etl tests/
+addopts = --cov=mozetl tests/
 
 [testenv]
 commands = pytest
-passenv = PYTHONPATH
```
harterrt: Do tests pass for you without passing the PYTHONPATH? Tox is complaining about not having access to pyspark.sql.

setup.py
```diff
@@ -1,21 +1,15 @@
...
-      version='0.1',
-      description='Python ETL jobs for Firefox Telemetry to be scheduled on Airflow.',
-      author='Ryan Harter',
-      author_email='harterrt@mozilla.com',
-      url='https://github.com/mozilla/python_etl.git',
-      packages=find_packages(exclude=['tests']),
-      install_requires=[
-          'python_moztelemetry',
-          ],
-      test_requires=[
```
harterrt: Looks like removing this section may be causing travis to fail.

tox.ini
jezdez: Ah, you're right!

setup.py
jezdez: Hm, they shouldn't, test_requires are only used when you run the tests via `python setup.py test`, which we don't. In python_moztelemetry there is a magic pytest running used for setuptools (see the `pytest-runner` entry in `setup_requires`). Which will then pick up the `test_requires` parameter.

jezdez:
@harterrt Fixed!

setup.py
harterrt: Ah, you're right. No change needed.

jezdez:
@harterrt \o/

PR #13, opened 2017-04-25
Stop SparkContext after finishing pulse job

harterrt:


harterrt:
Without this one-line change the airflow job will time-out.

PR #14, opened 2017-04-25
Bug 1359193 - Add topline historical backfill job

acmiyaguchi:
This job is a one off job that imports historical data from the v4 executive report scripts. This removes the `all` fields and writes the data into parquet compatible with `topline_summary`.

I chose to store the json version of the spark schema data-structure. This seems to be the most portable way to schema's across jobs. The schema is required for generating test input data, and for writing the data with the correct `topline_summary` schema.


mozetl/topline/historical_backfill.py
```diff
@@ -0,0 +1,119 @@
...
+    df = df.select(*[F.col(f.name).cast(f.dataType).alias(f.name)
+                     for f in topline_schema.fields])
+
+    # Use the same parititoning scheme as topline_summary
+    df.write.partitionBy('report_start').parquet(path)
+
+
+@click.command()
+@click.argument('s3_path')
+@click.argument('mode', type=click.Choice(['weekly', 'monthly']))
```
harterrt: Does `mode` change the way the data is processed or is it just a label on the created data?

setup.py
```diff
@@ -11,5 +11,6 @@
     packages=find_packages(exclude=['tests']),
     install_requires=[
         'python_moztelemetry',  # TODO: pin version
+        'click'
```
harterrt: Should we add an [entry point](http://click.pocoo.org/5/setuptools/) to setup.py?

tests/test_topline_historical_backfill.py
```diff
@@ -0,0 +1,144 @@
...
+    "crashes": "1",
+    "default": "1",
+    "google": "1",
+    "bing": "1",
+    "yahoo": "1",
+    "other": "1"
+}
+
+
+def generate_samples(snippets, base_sample):
```
harterrt: Can you comment on what snippets and base sample are?

mozetl/topline/historical_backfill.py
```diff
@@ -0,0 +1,119 @@
...
+        data = json.load(json_data)
+    return StructType.fromJson(data)
+
+
+# Generate module level schemas
+historical_schema = schema_from_json('historical_schema.json')
+topline_schema = schema_from_json('topline_schema.json')
+
+
+def create_s3_path(bucket, prefix, protocol="s3"):
```
mreid-moz: This function's name is confusing - it sounds like it might be creating something on S3, and it supports non-S3 paths.

Maybe something like `format_output_path(base, prefix, protocol="s3")` would be clearer.

tests/test_topline_historical_backfill.py
```diff
@@ -0,0 +1,144 @@
...
+    return samples
+
+
+def snippets_to_df(spark, snippets, base_sample, schema):
+    samples = generate_samples(snippets, base_sample)
+    jsonRDD = spark.sparkContext.parallelize(samples)
+    return spark.read.json(jsonRDD, schema=schema)
+
+
+# does not include all rows from original data
```
harterrt: This comment made sense only once I understood what was going on :). Do you mind quoting "all" to differentiate between 'all rows' and rows containing 'all'

mozetl/topline/historical_backfill.py
```diff
@@ -0,0 +1,119 @@
...
+    return "{}://{}/{}".format(bucket, prefix, protocol)
```
mreid-moz: params should be reordered to `protocol, bucket, prefix` here

mozetl/topline/historical_backfill.py
```diff
@@ -0,0 +1,119 @@
...
+              (F.col('channel') != 'all'))
+        .withColumn('report_start', F.date_format('date', 'yyyyMMdd'))
+    )
+
+    # Cast all elements from the csv file. Assume both schemas are flat.
+    df = df.select(*[F.col(f.name).cast(f.dataType).alias(f.name)
+                     for f in topline_schema.fields])
+
+    # Use the same parititoning scheme as topline_summary
+    df.write.partitionBy('report_start').parquet(path)
```
mreid-moz: Will this write a single file per partition? The output should be tiny, so we should make sure to coalesce to as few output files as possible.

mozetl/topline/historical_backfill.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+@click.command()
+@click.argument('s3_path')
```
mreid-moz: Can we give this argument a more descriptive name to differentiate it from the output S3 info? Something like "source_s3_path" or "historical_s3_path"?

mozetl/topline/topline_schema.json
```diff
@@ -0,0 +1,109 @@
...
+      "nullable":true,
+      "type":"long"
+    },
+    {
+      "metadata":{
+
+      },
+      "name":"report_start",
+      "nullable":true,
+      "type":"integer"
```
mreid-moz: Hm, I thought partition-based fields (stored in the S3 path) were treated as Strings. Good to know.

mozetl/topline/historical_backfill.py
acmiyaguchi: `mode` partitions the dataset, since there are two input csv files. `topline_summary` is laid out in s3 as follows: `telemetry-parquet/topline_summary/v1/mode=(weekly|monthly)/report_start=yyyyMMdd`.

setup.py
acmiyaguchi: I think that would be great to add, but it might be nice to get a few more scripts checked-in to get a feel for how multiple click commands integrate together. For the mean while, this script will be launched using `python -m mozetl.topline.historical_backfill <args>`.

mozetl/topline/historical_backfill.py
acmiyaguchi: Done.

mozetl/topline/historical_backfill.py
acmiyaguchi: I removed the protocol argument because it ends up not being used. I meant for the protocol flag to be set during mocking, but problems with recursive mocking lead to re-implement the function during mocking. 

This is a candidate for refactoring into a separate utility library.

mozetl/topline/historical_backfill.py
acmiyaguchi: It just so happens to write to a single file per partition column, but I'm not sure what set of commands would let me adjust partition sizes. Maybe `df.repartition(1, "report_start").write.partitionBy('report_start")` would do the trick? I think it's unnecessary in this case though.

mreid-moz: Agreed - as long as it writes a single file per partition in practice, that should be fine.

mozetl/topline/historical_backfill.py
acmiyaguchi: Done.

mozetl/topline/topline_schema.json
acmiyaguchi: I think it's because ToplineSummaryView doesn't have an explicit schema, but it ends up being treated as a string in Presto. I think the same thing is applicable to the churn dataset, and possibly(?) others. It could be potentially confusing down the road.

tests/test_topline_historical_backfill.py
acmiyaguchi: Done, I also renamed the test to avoid confusion.

tests/test_topline_historical_backfill.py
acmiyaguchi: Done.

This is a slightly modified version of the same function I used in the churn job, so it probably won't be the last time you see it. I think I might end up implementing tests in the same style for `topline_dashboard` and `churn`, so this would be a nice place to start refactoring after everything is in a working state.

mozetl/topline/topline_schema.json
acmiyaguchi: I'm pretty sure this should be a string field. MainSummaryView writes out the partition column in the same way, but `submission_date_s3` doesn't end up being read as an integer. I'm going to backfill the weekly topline data from airflow to double check this.

mreid-moz: Ok - I figured that when applying this schema, Spark coerced the string value from the S3 path into an integer.

mreid-moz:
Updates look good to me. I'll leave it up to you whether or not to modify the data type for the `report_start` field.

tests/test_topline_historical_backfill.py
harterrt: SG!

acmiyaguchi:
I've updated the datatype for `report_start`. Assuming `report_start` is an integer causes some issues with the schema down the line.

PR #18, opened 2017-04-27
 Bug 1357875 - Add script for generating topline dashboard data

acmiyaguchi:
This adds `topline_dashboard`, which is a replacement for the `v4_reformat.py` and `run.sh` of the original executive reporting pipeline. 

This PR is formatted in the following way:

1. Schema's are moved into the `topline` namespace to be shared across module
2. A skeleton of the script and a suit of tests are written
3. The script is completely implemented to pass all tests

This script adds moto to the list of test dependencies, which is useful for mocking s3 calls. In conjunction with the temporary directory fixture, pretty much everything you need (aside from moztelemetry) can be mocked with some work.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/18)


acmiyaguchi:
I'm also piggy-backing a few more commits to stop the SparkSession at the end of the function. This makes some tests fail since they share the same spark context. A fixture disables this SparkSession.stop function for the duration of the `historical_backfill` test.

mozetl/topline/topline_dashboard.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # find the file that was created
+    filepath = None
+    for name in os.listdir(path):
+        if name.endswith(".csv"):
+            filepath = os.path.join(path, name)
+            break
+
+    # name of the output key
+    key = "{}/v4-{}.csv".format(prefix, mode)
```
mreid-moz: Let's rename this to "topline-{weekly|monthly}.csv" to avoid conflicts with the old version of the output.

tests/test_topline_dashboard.py
```diff
@@ -0,0 +1,226 @@
...
+
+    conn = boto3.resource('s3', region_name='us-west-2')
+    conn.create_bucket(Bucket=bucket)
+
+    # dataframe is not reformatted, but not necessary
+    topline.write_dashboard_data(simple_df, bucket, prefix, 'weekly')
+
+    body = (
+        conn
+        .Object(bucket, prefix + '/v4-weekly.csv')
```
mreid-moz: Ditto re: naming "topline-X.csv"

tests/test_topline_dashboard.py
```diff
@@ -0,0 +1,226 @@
...
+    conn = boto3.resource('s3', region_name='us-west-2')
+    conn.create_bucket(Bucket=bucket)
+
+    # dataframe is not reformatted, but not necessary
+    topline.write_dashboard_data(simple_df, bucket, prefix, 'weekly')
+    topline.write_dashboard_data(multi_df, bucket, prefix, 'weekly')
+
+    body = (
+        conn
+        .Object(bucket, prefix + '/v4-weekly.csv')
```
mreid-moz: Ditto re: naming "topline-X.csv"

tests/test_topline_dashboard.py
```diff
@@ -0,0 +1,226 @@
...
+        prefix,
+        '--input_bucket', input_bucket
+    ]
+    result = runner.invoke(topline.main, args)
+    assert result.exit_code == 0
+
+    # read results using boto
+    body = (
+        conn
+        .Object(bucket, prefix + '/v4-monthly.csv')
```
mreid-moz: Ditto re: naming "topline-X.csv"

mozetl/topline/topline_dashboard.py
```diff
@@ -0,0 +1,197 @@
+""" Reformat Data for Topline Dashboard
+
+This script replaces `run.sh` and `v4_reformat.py` that generates data ready
+for dashboard consumption. The original system used a pipeline consisting of a
```
mreid-moz: Let's add a link to the dashboard @ `http://metrics.services.mozilla.com/firefox-dashboard/` in this comment and call out that the old data files were located in `data/v4-weekly.csv` and `data/v4-monthly.csv`. That might be helpful if we need to trace the data's origin in the future.

mozetl/topline/topline_dashboard.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # Note: The schema is applied to cast report_start to a string. The
+    # DataFrameReader interface has a `.schema()` that mysteriously fails
+    # in this context, so the schema is applied after reading the dataframe
+    # into memory.
+    topline_summary = (
+        spark.read
+        .parquet(input_path)
+        .rdd.toDF(topline_schema)
+        .where(F.col('report_start') == report_start)
```
mreid-moz: The dashboard output should include data for all time - I don't think we want to filter to a specific report period here. We may want to add optional min/max `report_start` values for debugging, but in general we want all weekly or monthly data in the final output csv.

If possible, please also add a test case to verify the above behaviour.

acmiyaguchi: I'm just going to remove report_start, since it doesn't look like its going to see much use. I think this was left-over from my initial sketch that was going to more closely follow the run.sh script.

mreid-moz: Sounds good

acmiyaguchi:
@mreid-moz  this is ready for another look.

mreid-moz:
Looks good, thanks! @harterrt I defer to you for any `python_mozetl`isms :)

tests/test_topline_dashboard.py
```diff
@@ -0,0 +1,225 @@
...
+
+import boto3
+import pytest
+from moto import mock_s3
+from mozetl.topline import topline_dashboard as topline
+from mozetl.topline.schema import historical_schema, topline_schema
+from pyspark.sql import SparkSession
+
+
+@pytest.fixture(scope="session")
```
acmiyaguchi: I'll address this in another PR, since it'll be touching a couple of different files.

acmiyaguchi:
Thanks for the review @mreid-moz, @harterrt :+1:

PR #27, opened 2017-05-04
Bug 1357875 - Use toLocalIterator instead of writing directly to `file://` via spark

acmiyaguchi:
There are certain permission issues when deploying topline_dashboard due to permission errors with the spark executor. One alternative is to write the csv files to hdfs and copy over to local disk. This is not necessarily an elegant solution to the problem, so the tried solution is to use the collect (as long as this dataset stays within the size of the driver).

This patch works in practice, and was used to generate reports under `net-mozaws-prod-us-west-2-pipeline-analysis/amiyaguchi/topline-dashboard-data/`

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/27)

### DIFF
--- a/mozetl/topline/topline_dashboard.py
+++ b/mozetl/topline/topline_dashboard.py
@@ -28,7 +28,7 @@ import click
 
 from pyspark.sql import SparkSession, functions as F
 from mozetl.topline.schema import topline_schema, historical_schema
-
+from mozetl import utils
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -122,16 +122,9 @@ def write_dashboard_data(df, bucket, prefix, mode):
 
     # create a temporary directory to dump files into
     path = tempfile.mkdtemp()
+    filepath = os.path.join(path, 'temp.csv')
 
-    # write dataframe to working directory
-    df.repartition(1).write.csv(path, header=True, mode="overwrite")
-
-    # find the file that was created
-    filepath = None
-    for name in os.listdir(path):
-        if name.endswith(".csv"):
-            filepath = os.path.join(path, name)
-            break
+    utils.write_csv(df, filepath)
 
     # name of the output key
     key = "{}/topline-{}.csv".format(prefix, mode)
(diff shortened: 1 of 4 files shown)


mozetl/topline/topline_dashboard.py
```diff
@@ -122,16 +122,17 @@ def write_dashboard_data(df, bucket, prefix, mode):
...
+
+    with open(filepath, 'wb') as f:
+        f.write(','.join(df.columns) + '\n')
+        rows = df.collect()
+        for row in rows:
+            try:
+                f.write(','.join([unicode(c) for c in row]))
+                f.write('\n')
+            except UnicodeEncodeError:
+                logger.exception("exception while processing csv")
```
mreid-moz: I don't know if it's a good idea to potentially drop rows with just a logged message. Does this happen in practice? If so, under what circumstances?

acmiyaguchi: This doesn't happen in practice because the topline_summary is generally well formed. I don't don't think there are malformed strings in the original data.

mreid-moz: Ok, can you confirm that writing bare 'strings' (rather than u'unicode' strings) works ok in combination with unicode field values?

It might be worthwhile to factor out some "write this dataframe to csv" functionality and add some tests. handling all the unicode/utf8 corner cases in python2 always seems to be fraught with peril.

acmiyaguchi:
The source of the failing tests is from a bad interaction of pyspark rdd methods with moto. I've patched the offending methods.

I've also put the csv writer method into a utility module. I still need to write a few tests that verify the behavior with unicode/utf-8.

mozetl/utils.py
```diff
@@ -0,0 +1,14 @@
+import csv
+import pdb
+
+
+def write_csv(dataframe, path):
+    """ Write a dataframe to local disk. """
+
+    # pdb.set_trace()
```
mreid-moz: Can this be removed (along with the import above)?

acmiyaguchi:
@mreid-moz RFAL

PR #28, opened 2017-05-08
Bug 1361186: Port TxP Mau Dau job from ipynb

sunahsuh:
This ports over the TxP Mau Dau job from a python notebook to a python script so we can schedule it with airflow.

In the process of porting, however, I also made some significant modifications to the format and process, simplifying both. As I mentioned in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1361186), I think this level of change might warrant writing to a different location and allowing query owners to move over their queries one by one.

I realize I have no tests -- they'd be a nice-to-have, but I don't think this job warrants that effort at the moment.

cc @ilanasegall

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/28)

### DIFF
--- a/mozetl/testpilot/txp_mau_dau.py
+++ b/mozetl/testpilot/txp_mau_dau.py
@@ -1,0 +1,114 @@
+from os import environ
+import logging
+from datetime import datetime as dt, date, timedelta
+import requests
+from pyspark.sql.functions import col, countDistinct, lit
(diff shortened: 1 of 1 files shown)


sunahsuh:
RFAL -- I also updated the spark session handling a bit

PR #32, opened 2017-05-11
Cache and materialize a dataframe before writing to disk

acmiyaguchi:
Materializing the dataframe/rdd can take a while, which often causes
toLocalIterator to timeout. [1] This is solved on newer versions of
pyspark, but can be worked around by caching and materializing the
dataframe before calling toLocalIterator.

[1] https://issues.apache.org/jira/browse/SPARK-18281

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/32)

### DIFF
--- a/mozetl/utils.py
+++ b/mozetl/utils.py
@@ -1,8 +1,19 @@
 import csv
+import logging
+
+
+logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
@mreid-moz bump, this should be the first PR to merge to fix tests.

Issue #34, opened 2017-05-16
Failing Tests


harterrt:
Looks like the tests are [failing on master](https://travis-ci.org/mozilla/python_mozetl/builds/231195163). Anthony, looks like it's the topline dashboard (or some change in the moto module). Can you take a look?

acmiyaguchi:
I made a [pull request to fix this upstream](https://github.com/spulec/moto/pull/950). I can add mock as a dependency and remove some of the monkey patching I wrote to get around bad library interactions.

acmiyaguchi:
There's still one failure left, which is probably related to toLocalIterator.

PR #36, opened 2017-05-16
Fix issue #34 caused by missing mock dependency

acmiyaguchi:
The update to mock 1.0.0 introduces a missing dependency, but also fixes
an older issue of mock not interacting well with pyspark. This removes
monkeypatching for older tests.

PR #32 should be addressed to fix tests that are failing due to a lack of rdd materialization in this pull request.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/36)

### DIFF
--- a/tests/test_topline_dashboard.py
+++ b/tests/test_topline_dashboard.py
@@ -59,25 +59,16 @@ def snippets_to_df(spark, snippets, base_sample, schema):
     return spark.read.json(jsonRDD, schema=schema)
 
 
-def patch_iterator(monkeypatch, rows):
-    # toLocalIterator does not work when mocking with moto. Patch this
-    # to return the right result
-    def mock_iterator(self):
-        return rows
-    monkeypatch.setattr("pyspark.sql.DataFrame.toLocalIterator", mock_iterator)
-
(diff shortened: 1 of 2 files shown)


tests/test_topline_dashboard.py
```diff
@@ -59,20 +59,11 @@ def snippets_to_df(spark, snippets, base_sample, schema):
...
-def patch_iterator(monkeypatch, rows):
-    # toLocalIterator does not work when mocking with moto. Patch this
-    # to return the right result
-    def mock_iterator(self):
-        return rows
-    monkeypatch.setattr("pyspark.sql.DataFrame.toLocalIterator", mock_iterator)
-
-
 @pytest.fixture()
 def simple_df(spark, monkeypatch):
```
harterrt: Can you drop the monkeypatch parameter as well? Here and below.

mreid-moz: The `monkeypatch` arg is no longer used. Can it be removed?

acmiyaguchi:
@harterrt @mreid-moz RFAL

PR #43, opened 2017-05-25
start mozetl/shield/containers.py from testpilot

groovecoder:
The only changes are:
```
lambda ping: ping['payload/test'] in ["@testpilot-containers", "@shield-study-containers"]
```
and:
```
's3n://telemetry-parquet/harter/containers_testpilottest/v2'
```

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/43)


groovecoder:
I assume if these changes are correct, I'll create new/duplicate queries on STMO. Will the `'s3n://telemetry-parquet/harter/containers_testpilottest/v2'` automatically create a new `containers_testpilottest_v2` table? So I can just update the new/duplicate queries to use that table and (initially) filter out the `@testpilot-containers` records for a shield-only dashboard?

mozetl/shield/containers.py
```diff
@@ -0,0 +1,34 @@
+from pyspark.sql.types import StringType, LongType
+
+from ..basic import convert_pings, DataFrameConfig
+from .utils import testpilot_etl_boilerplate
+
+
+def transform_pings(sqlContext, pings):
```
harterrt: The logic change looks right, but it would be nice if we could cut down on the code duplication. It's fine if this job lives in testpilot/containers.py (even though this is a shield study).

mozetl/shield/containers.py
```diff
@@ -0,0 +1,34 @@
...
+            ("hiddenContainersCount", "payload/payload/hiddenContainersCount", None, LongType()),
+            ("shownContainersCount", "payload/payload/shownContainersCount", None, LongType()),
+            ("totalContainersCount", "payload/payload/totalContainersCount", None, LongType()),
+            ("totalContainerTabsCount",
+             "payload/payload/totalContainerTabsCount", None, LongType()),
+            ("totalNonContainerTabsCount",
+             "payload/payload/totalNonContainerTabsCount", None, LongType()),
+            ("pageRequestCount", "payload/payload/pageRequestCount", None, LongType()),
+            ("test", "payload/test", None, StringType()),
+        ], lambda ping: ping['payload/test'] in ["@testpilot-containers", "@shield-study-containers"])
```
harterrt: I know we've discussed this before, but I can't remember the conclusion. Do we actually want shield and testpilot data in one table? Will we be comparing testpilot and shield users or will we focus our analysis on the differences between experiment and control within the shield study?

mozetl/shield/containers.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+
+etl_job = testpilot_etl_boilerplate(
+    transform_pings,
+    's3n://telemetry-parquet/harter/containers_testpilottest/v2'
```
harterrt: You should probably choose a new location for this table. Storing this table as v2 will cause [parquet2hive](https://github.com/mozilla/parquet2hive) to load the new shield table into containers_testpilottest, so only the combined shield/txp table will be available from redash. 

Maybe `s3n://telemetry-parquet/harter/containers_txp_shield/v1` ?

mozetl/shield/containers.py
groovecoder: I think we decided to leave the existing table as-is, and create a new table that has both kinds of pings in it. That way we can do both: compare testpilot & shield users to each other, and do some focused analysis on the behaviors of the different shield variant users.

mozetl/shield/containers.py
groovecoder: I'm not sure how the plumbing of mozetl works ... can I simply put this at the bottom?
```
testpilot_etl_job = testpilot_etl_boilerplate(
      transform_testpilot_pings,
      's3n://telemetry-parquet/harter/containers_testpilottest/v1'
  )

shield_etl_job = testpilot_etl_boilerplate(
      transform_shield_pings,
      's3n://telemetry-parquet/harter/containers_shield/v1'
  )
```

Or can there be only 1 `etl_job` symbol in the module?

mozetl/shield/containers.py
groovecoder: Makes sense.

mozetl/shield/containers.py
harterrt: Perfect, SGTM

mozetl/shield/containers.py
harterrt: That should work. [This notebook](https://gist.github.com/harterrt/fb9cdc217da4e0410556c9190132d0a5) is scheduled to run every day on ATMO. It would be trivial to update the notebook to run both the `testpilot_etl_job` and `shield_etl_job` instead. 

The other option would be wrapping these two function in a third `etl_job` which would require no modification elsewhere.

groovecoder:
Okay, I addressed the comments here by:

1. Moving/consolidating all code back into the single `mozetl/testpilot/containers.py` module
2. defining a wrapper `etl_job` function in the module, which calls `testpilot_etl_job` and `shield_etl_job`
3. Re-naming the `s3_path` for shield to `containers_shield/v1`

Have another look?

groovecoder:
If I merge, will this start running immediately? Do we need to set anything else up, or can I merge and we'll see what happens?

harterrt:
Go ahead and merge. It should run this afternoon, but a merge will not trigger an immediate run.

groovecoder:
@harterrt - did the job run okay yesterday?

harterrt:
@groovecoder, The job failed yesterday but passed after we submitted [PR 47](https://github.com/mozilla/python_mozetl/pull/47)

PR #48, opened 2017-05-31
Filter followon searches from search daily job

harterrt:
I am still figuring out how we can test this on real data without causing a failure when the airflow job runs.

https://bugzilla.mozilla.org/show_bug.cgi?id=1368089

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_mozetl/48)

### DIFF
--- a/mozetl/search/daily_search_rollups.py
+++ b/mozetl/search/daily_search_rollups.py
@@ -57,6 +57,11 @@ TEMP_TABLE_TEMPLATE = 'rollup_temp_{}'
 # default search engine (or locale or geo), or a profile may be shared by
 # multiple users on a day (causing all the other fields to vary if the
 # users are on widely distributed machines).
+#
+# Restrict to a whitelist of search_source's to avoid double counting while
+# we test our new search_count telemetry developed in:
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1367554
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1368089
 SELECT_ROLLUP_TEMPLATE = """
 SELECT
   submission_date,
@@ -72,6 +77,8 @@ FROM
   {}
 WHERE
   ((search_count > -1) OR (search_count is null))
+  AND search_source IN ('searchbar', 'urlbar', 'abouthome', 'newtab',
+                        'contextmenu', 'system', 'NO_SEARCHES')
 GROUP BY
   submission_date, country, search_provider, default_provider, locale, distribution_id
 """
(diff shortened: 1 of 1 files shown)


harterrt:
Just tested this on data from 2017-05-29 with sample_ids in range(10). The data is identical before and after, which is what we hope to see. I don't have any way to test for upcoming followon keys until we refactor this job to be testable. 

Adding Dave as a reviewer to make sure my key filter format makes sense.

acmiyaguchi:
Another +1 for the whitelist approach, its probably more resilient to change. It might be useful to make a comment with a short description of this change and a few examples of keys to be filtered out, as notes to use during a refactor.

mozetl/search/daily_search_rollups.py
```diff
@@ -129,6 +130,7 @@ def roll_up_searches(spark, frame):
         "distribution_id",
         "default_provider",
         "search_counts.engine as search_provider",
+        "search_counts.source as search_source",
```
acmiyaguchi: A comment about filtering out search sources not related to search engines. An example of ignored source (`filteron%`).

harterrt: See above. The bug links should be the most useful resource for future engineers.

harterrt:
Moved to a whitelist, added a comment, and reran over 3% of one day's worth of main_summary data to verify nothing in the rollup changes.

RFAL.

mozetl/search/daily_search_rollups.py
```diff
@@ -72,6 +77,7 @@ def add_profile_day(frame):
   {}
 WHERE
   ((search_count > -1) OR (search_count is null))
+  AND search_source IN ('searchbar', 'urlbar', 'abouthome', 'newtab', 'contextmenu', 'system')
```
dzeber: I think we also need `'NO_SEARCHES'` in here to pull in the rows from the `unwrapped_nulls` table.

harterrt: Done, nice catch!

harterrt:
OK - looks like my manual tests passed without good reason. I've added "NO_SEARCHES" to the filter and verified that the version of the code missing this source is broken and the current version matches production.

I'll submit in the morning to make sure nothing new pops up :)

harterrt:
Submitting on Friday and crossing my fingers. We already have to backfill this job for 2017-05-31 and 2017-05-30 for a separate failure, so it will be good to test this job against that data.

PR #52, opened 2017-06-15
Bug 1364530 - Migrate Search ETL

acmiyaguchi:
[Bug 1364530 - Migrate Search ETL job to python_mozetl](https://bugzilla.mozilla.org/show_bug.cgi?id=1364530)

This PR completes the rest of the bug above.

* Runs on a smaller number of nodes (1-3 seems acceptable)
* Airflow friendly
* Generates daily and monthly rollups
* Output is backwards compatible with old datasets
* Adds tests for ETL logic and CLI/handoff

I need to validate the results of this job again against daily and monthly rollups. There seems to be to be a small margin of error that is consistent with the old daily rollup when the sharding/collect logic is removed.

Also ignore the noise generated from the first commit, that should be handled in #51.


acmiyaguchi:
@SamPenrose I can't seem to find you on the list of reviewers, I'd appreciate the feedback on this job though.

SamPenrose:
tox invokes flake8, which has one complaint about mozetl/search/daily_search_rollups.py:326 and many about the tests.

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
-    report(date, path, bad)
+@click.argument('start_date')
+@click.argument('mode', type=click.Choice(['daily', 'monthly']))
+@click.argument('bucket')
+@click.argument('prefix')
+@click.option('--input_bucket',
+              default='telemetry-parquet',
+              help='Bucket of the input dataset')
+@click.option('--input_prefix',
+              default='main_summary/v4',
```
SamPenrose: Question: should this be in a configuration constant? In particular, should the version # be in a constant?

tests/test_search_rollup.py
```diff
@@ -0,0 +1,285 @@
+import functools
```
SamPenrose: Overall these tests look good. Nit: they don't address the relationship between NULL and aggregation, which has been the dominant source of bugs.

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
-    msg['To'] = owner
-    msg['From'] = owner
-    smtp = smtplib.SMTP('localhost')
-    smtp.sendmail(owner, owner, msg.as_string())
+
+    # persist the dataframe to disk
+    logging.info("Writing dataframe to {}/{}".format(bucket, key))
+    utils.write_csv_to_s3(dataframe.select(select_expr), bucket, key, header=False)
+
+    csv_paths = get_csv_locations(bucket, key)
```
SamPenrose: Nit: I would be inclined to define the paths we will write in one place, rather than relying on retrieving them later ... particularly from a filesystem.

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
+    return (
+        spark.read
+        .option("mergeSchema", "true")
+        .parquet(path)
+        .where((F.col("submission_date_s3") >= ds_start) &
+               (F.col("submission_date_s3") < ds_end))
+    )
+
+
+def get_last_manifest_version(bucket, prefix, pattern):
```
SamPenrose: Comment: Thanks for so carefully improving my manifest-iteration logic. The next time we refactor I'd be inclined to rip all this out and come up with some simpler way of creating unique basenames.

tests/test_search_rollup.py
acmiyaguchi: I'll go ahead and add some tests that target NULL values. That seems pretty important to validate.

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
         "country",
         "locale",
         "distribution_id",
-        "default_search_engine as default_provider",
-        "'NO_SEARCHES' as search_provider",
-        "'NO_SEARCHES' as search_source",
-        "0 as search_count"
+        "default_provider",
+        "profile_count",
+        "profile_share",    # only for weekly
```
SamPenrose: Nit: "only for daily"

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
-        "default_provider",
-        "search_counts.engine as search_provider",
-        "search_counts.source as search_source",
-        "search_counts.count as search_count")
-    unwrapped_nulls = nulls_frame.selectExpr(
-        "submission_date",
-        "profile_day",
+    ]
+
+    defaults = {
```
SamPenrose: Thank you for extracting these!

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
+    :main_summary DataFrame: source table
+    :mode str: `daily` or `monthly`
+    """
 
-    exploded = frame.selectExpr(
-        "submission_date",
-        "profile_day",
+    # take a subset of the original dataframe
+    columns = [
+        F.col("client_id").alias("profile"),
```
SamPenrose: Question: why the rename? Do we do so elsewhere?

SamPenrose:
Overall this is great work and a big improvement over what I handed you. Thank you very much. f+, with feeling. Nits and questions inline; none blocking.

mozetl/search/daily_search_rollups.py
acmiyaguchi: I think the proper way to address this is to have an up to date list of all logical tables leaving the low levels locations opaque to the user, like the hive metastore. 

I lean toward leaving this as is, since its consistent with other jobs that I have written. I think there is room for improvement here, but it should probably be more than a `MAIN_SUMMARY_PREFIX`/`MAIN_SUMMARY_VERSION` constant. I like the idea of having a centralized store with information about the location and version of all datasets, but it's also a lot of work to keep it up to date and robust.

mozetl/search/daily_search_rollups.py
acmiyaguchi: I also feel the same, I think this can be refactored out since I moved from using `dataframe.write.csv` to `utils.write_csv_to_s3`. This was originally needed to specify the location of the csv files written by spark, which originally could have been in multiple parts.

mozetl/search/daily_search_rollups.py
acmiyaguchi: I'd rather make changes now than later, since I'm already in a refactoring mode. I tried to keep most of the functionality on par with the original script.

But, do we actually need to iterate through multiple versions of the manifest?

The csv files are written to a deterministic location with a fairly stable naming convention. The nature of the roll-up leaves the csv files fairly small in size, so they doesn't need to be split across multiple files. The backfill will also overwrite the original csv file, meaning that the various iterations of the manifest will be pointing to the same file. I'm all for removing the logic for generating iterative versions of the manifest (and possibly the manifest itself).

mozetl/search/daily_search_rollups.py
acmiyaguchi: :+1:

mozetl/search/daily_search_rollups.py
SamPenrose: Right now the protocol with IT's ingestion script is: new manifest file triggers a new ingestion. As long as we conform to that (i.e. don't reuse a path and expect it to be ingested into Vertica), I am fine with deleting logic.

mozetl/search/daily_search_rollups.py
SamPenrose: +1 on leave as is for now.

mozetl/search/daily_search_rollups.py
acmiyaguchi: I don't really want to needlessly break scripts, the handoff piece seems to be out of the scope of this bug anyways. I'm happy to leave this as it is.

acmiyaguchi:
@mreid-moz RFAL

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
+    other fields to vary if the users are on widely distributed
+    machines).
+
+    :dataframe DataFrame: slightly processed data
+    :attributes list[str]: columns to group over
+    :mode str: `daily` or `monthly`, used to remove `profile_shares`
+    """
+    metrics = {
+        "search_count": F.sum("search_count").alias("search_count"),
+        "profile": F.countDistinct("profile").alias("profile_count"),
```
mreid-moz: I am confused by the change from `profile_day` to `profile` (which, if I understand correctly, is just renamed `client_id`). Isn't this going to affect the results? It doesn't seem like it would make any difference in "daily" mode, but what about "monthly"?

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
-        PROFILE_COLUMN
+    # return their union
+    return exploded_searches.union(null_searches)
+
+
+def with_shares(dataframe):
+    """ Assign each client a weight for the contribution toward the
+    rollup aggregates.
+
+    Given a single profile_day with N permutations of
```
mreid-moz: The code no longer appears to use `profile_day` - this comment should be updated accordingly.

Also, is there test coverage for this logic?

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
-      1.0/count(*) AS profile_share
-    FROM
-      {}
-    GROUP BY profile_day)
-  share_table
-WHERE
-  {}.profile_day = share_table.profile_day
-"""
+def rollup_searches(dataframe, attributes, mode):
+    """ Gather all permutations of submission_date, country,
```
mreid-moz: Please update this comment to remove the explicit column list and say something like "all permutations of columns specified by the `attributes` arg"

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
+    search_provider, default_provider, locale, distribution_id and
+    count:
+    1) How many searches fall into each permutation? -> search_count
+    2) How many profiles fall into each bucket -> profile_count
```
mreid-moz: please mention that we're counting *unique* profiles

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
+    :version int: version of the rollup
+    :start_ds str: yyyymmdd
+    """
+
+    # format the save location of the data
+    start_date = arrow.get(start_ds, "YYYYMMDD")
+
+    # select the relevant fields
+    select_expr = [
+        "ROLLUP_DATE_PLACEHOLDER",
```
mreid-moz: I think it would be slightly simpler to begin with this value set to `F.lit(start_date.format("YYYY-MM-DD"))`, then update it only for the `mode=monthly` case.

mozetl/search/daily_search_rollups.py
```diff
@@ -1,309 +1,371 @@
...
+        if obj.key.endswith(".csv"):
+            csv_files.append("s3://{}/{}".format(bucket, obj.key))
+    return csv_files
+
+
+def write_manifest(bucket, prefix, mode, version, start_ds, csv_paths, retry_max=10):
+    """ Write a manifest file with the location of the daily rollup files.
+
+    A manifest enumerates the locations of csv files that contain the rollup.
+    This might be useful if the file is partitioned across multiple files.
```
mreid-moz: Please add a comment indicating that the manifest is needed as part of the subsequent Vertica integration per [Sam's comment](https://github.com/mozilla/python_mozetl/pull/52/files#r122309470)

mozetl/search/daily_search_rollups.py
acmiyaguchi: The monthly report takes the distinct profiles seen in a month, which is why the change from `profile_day` to `profile` was made. The logic between the two jobs are identical aside from the `profile_shares`.

mreid-moz: Ok cool, as long as it's working as designed, sgtm.

mozetl/search/daily_search_rollups.py
acmiyaguchi: Yes, see `test_single_client_shares_multiple_countries`

tests/test_search_rollup.py
```diff
@@ -8,7 +8,7 @@
 from pyspark.sql.types import StructType, StructField, StringType, ArrayType, \
     LongType
 
-from mozetl.search import daily_search_rollups as search_rollups
+from mozetl.search import search_rollups
```
mreid-moz: We'll need a corresponding PR over in `telemetry-airflow` to run this using the new name:
https://github.com/mozilla/telemetry-airflow/blob/master/jobs/search_rollups.sh#L9

mreid-moz: See https://github.com/mozilla/telemetry-airflow/pull/137

PR #54, opened 2017-06-21
Bug 1370522 - Port official engagement ratio heuristic to this repo

SamPenrose:
In addition to the unit tests I ran against main_summary and it matched the results of the existing notebook.

### DIFF
--- a/mozetl/maudau/maudau.py
+++ b/mozetl/maudau/maudau.py
@@ -1,0 +1,235 @@
+"""
+Calculate Firefox monthly- and daily-active users.
+
+main_summary is updated daily; re-runs should be idempotent between
+updates.
+C.f. https://bugzilla.mozilla.org/show_bug.cgi?id=1370522,
+     https://bugzilla.mozilla.org/show_bug.cgi?id=1240849
+"""
+import csv
+import datetime as DT
+import logging
+import os
+
+from pyspark.sql import SparkSession
+import boto3
+import botocore
+import click
+from boto3.s3.transfer import S3Transfer
+from moztelemetry.standards import (
+    filter_date_range,
+    count_distinct_clientids
+)
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+DEVELOPMENT = False
+
+STORAGE_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
+STORAGE_SUB_DIR = 'spenrose/maudau' if DEVELOPMENT else 'mreid/maudau'
+MAUDAU_ROLLUP_BASENAME = "engagement_ratio.csv"
+MAUDAU_SNAPSHOT_TEMPLATE = "engagement_ratio.{}.csv"
+DASHBOARD_BUCKET = "net-mozaws-prod-metrics-data"
+DASHBOARD_KEY = "firefox-dashboard/{}".format(MAUDAU_ROLLUP_BASENAME)
+
+ACTIVITY_SUBMISSION_LAG = DT.timedelta(10)
+# MONTH as in "monthly active"
+MONTH = 28
+
+
+def format_as_submission_date(date):
+    return DT.date.strftime(date, "%Y%m%d")
+
+
+def parse_as_submission_date(date_string):
+    return DT.datetime.strptime(date_string, "%Y%m%d").date()
+
+
+def format_spark_path(bucket, prefix):
+    return "s3://{}/{}".format(bucket, prefix)
+
+
+def get_rollup_s3_paths(basename):
+    return (STORAGE_BUCKET, os.path.join(STORAGE_SUB_DIR, basename))
+
+
+def generate_filter_parameters(end_date, days_back):
+    d = {}
+    min_activity_date = end_date - DT.timedelta(days_back)
+    d['min_activity_iso'] = min_activity_date.isoformat()
+    d['max_activity_iso'] = (end_date + DT.timedelta(1)).isoformat()
+
+    d['min_submission_string'] = format_as_submission_date(min_activity_date)
+    max_submission_date = end_date + ACTIVITY_SUBMISSION_LAG
+    d['max_submission_string'] = format_as_submission_date(max_submission_date)
+    return d
+
+
+def count_active_users(frame, end_date, days_back):
+    """
+    Tally active users according to externally-defined heuristics,
+    specifically client-reported subsession_start_date.
+
+    :frame DataFrame(): conforming to main_summary schema
+    :end_date DT.date(): the last day on which to count activity.
+    :days_back int: how far back to go; 0 for daily and 28 for monthly.
+
+    Note that return values are not stable over time as activity trickles
+    in with long submission latencies.
+
+    See https://bugzilla.mozilla.org/show_bug.cgi?id=1240849
+    """
+    params = generate_filter_parameters(end_date, days_back)
+
+    filtered = filter_date_range(
+        frame,
+        frame.subsession_start_date,
+        params['min_activity_iso'],
+        params['max_activity_iso'],
+        frame.submission_date_s3,
+        params['min_submission_string'],
+        params['max_submission_string'])
+    return count_distinct_clientids(filtered)
+
+
+def parse_last_rollup(basename, start_date=None):
+    """
+    Extract the valid old counts from basename and return them with the
+    first date that needs to be re-counted.
+    """
+    start_date = start_date or DT.date.today()
+    since = start_date - ACTIVITY_SUBMISSION_LAG
+    carryover = []
+    with open(basename) as f:
+        reader = csv.DictReader(f)
+        last_day = None
+        for row in reader:
+            day = parse_as_submission_date(row['day'])
+            if day >= since:
+                break
+            if last_day is not None:
+                if (day - last_day) > DT.timedelta(1):
+                    since = last_day + DT.timedelta(1)
+                    break
+            last_day = day
+            carryover.append(row)
+    return since, carryover
+
+
+def get_last_rollup(transferer):
+    '''
+    Always chose the latest rollup, even if start_date is old.
+    '''
+    basename = MAUDAU_ROLLUP_BASENAME
+    key = os.path.join(STORAGE_SUB_DIR, basename)
+    try:
+        transferer.download_file(
+            STORAGE_BUCKET,
+            key,
+            basename)
+    except botocore.exceptions.ClientError as e:
+        # If the file wasn't there, that's ok. Otherwise, abort!
+        if e.response['Error']['Code'] != "404":
+            raise e
+        else:
+            msg = "Did not find an existing file at '{}'".format(basename)
+            logger.info(msg)
+        return None
+    return basename
+
+
+def generate_counts(frame, since, until=None):
+    '''
+    A thin wrapper around moztelemetry counting functions.
+
+    :frame main_summary or an isomorphic subset
+    :since DT.date()
+    :until DT.date() (unit test convenience)
+    '''
+    cols = [frame.client_id.alias('clientId'),
+            'subsession_start_date', 'submission_date_s3']
+    narrow = frame.select(cols)
+    updates = []
+    today = DT.date.today()
+    generated = format_as_submission_date(today)
+    start = since
+    until = until or today
+    while start < until:
+        dau = count_active_users(narrow, start, 0)
+        mau = count_active_users(narrow, start, MONTH)
+        day = format_as_submission_date(start)
+        d = {'day': day, 'dau': dau, 'mau': mau, 'generated_on': generated}
+        updates.append(d)
+        start += DT.timedelta(1)
+    return updates
+
+
+def write_locally(results):
+    '''
+    :results [{'day': '%Y%m%d', 'dau': <int>,
+              'mau': <int>, 'generated_on': '%Y%m%d'}, ...]
+    '''
+    publication_date = format_as_submission_date(DT.date.today())
+    basename = MAUDAU_SNAPSHOT_TEMPLATE.format(publication_date)
+    cols = ["day", "dau", "mau", "generated_on"]
+    with open(basename, 'w') as f:
+        writer = csv.DictWriter(f, cols)
+        writer.writeheader()
+        writer.writerows(results)
+    return basename
+
+
+def publish_to_s3(transferer, bucket, prefix, basename):
+    '''
+    Write the file twice to storage (once dated, once as "latest")
+    and once to the production dashboard.
+    '''
+    dated_key = os.path.join(prefix, basename)
+    transferer.upload_file(basename, bucket, dated_key)
+    latest_key = os.path.join(prefix, MAUDAU_ROLLUP_BASENAME)
+    transferer.upload_file(basename, bucket, latest_key)
+    if DEVELOPMENT:
+        return
+    transferer.upload_file(basename, DASHBOARD_BUCKET, DASHBOARD_KEY)
+
+
+@click.command()
+@click.option('--input_bucket',
+              default='telemetry-parquet',
+              help='Bucket of the input dataset')
+@click.option('--input_prefix',
+              default='main_summary/v4',
+              help='Prefix of the input dataset')
+@click.option('--output_bucket',
+              default=STORAGE_BUCKET,
+              help='Bucket of the output dataset')
+@click.option('--output_prefix',
+              default=STORAGE_SUB_DIR,
+              help='Prefix of the output dataset')
+def main(input_bucket, input_prefix, output_bucket, output_prefix):
+    s3client = boto3.client('s3', 'us-west-2')
+    transferer = S3Transfer(s3client)
+    last_rollup_basename = get_last_rollup(transferer)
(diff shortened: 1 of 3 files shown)


tests/test_maudau.py
```diff
@@ -0,0 +1,127 @@
...
+    values = [
+        ('a', '2017-05-01', '20170508'),
+        ('b', '2017-05-01', '20170501'),
+        ('c', '2017-05-01', '20170510'),
+        ('a', '2017-05-02', '20170503'),
+        ('b', '2017-05-03', '20170503'),
+        ('b', '2017-05-04', '20170511'),
+        ('a', '2017-05-05', '20170507'),
+    ]
+    rows = map(json.dumps, [dict(zip(cols, tup)) for tup in values])
```
acmiyaguchi: Nit: I think you should be able to create a dataframe by using `spark.createDataFrame(list_of_dicts, schema)`, instead of going through a json -> rdd -> dataframe transformation.

tests/test_maudau.py
```diff
@@ -0,0 +1,127 @@
...
+    try:
+        os.chdir(tempdir)
+        M.write_locally(results)
+        output = open(basename).read()
+        assert output == expected
+    finally:
+        os.chdir(cwd)
+
+
+if __name__ == '__main__':
```
acmiyaguchi: Adding the test functions here shouldn't be necessary if everything is set up correctly (although it is pretty difficult to set up the testing environment correctly). Pytest should be able to find these tests automatically, and only these tests can be run by specifying it in the commandline in the project root: `pytest -k test_maudau.py`

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@click.command()
+@click.option('--input_bucket',
+              default='telemetry-parquet',
+              help='Bucket of the input dataset')
+@click.option('--input_prefix',
+              default='main_summary/v4',
+              help='Prefix of the input dataset')
+def main(input_bucket, input_prefix):
```
acmiyaguchi: Having the output bucket and prefix would be useful for testing on airflow -- the staging instance can specify a testing bucket so it doesn't affect the production files.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+DASHBOARD_BUCKET = "net-mozaws-prod-metrics-data"
+DASHBOARD_KEY = "firefox-dashboard/{}".format(MAUDAU_ROLLUP_BASENAME)
+
+today = DT.date.today()
+yesterday = today - DT.timedelta(1)
+ACTIVITY_SUBMISSION_LAG = DT.timedelta(10)
+# MONTH as in "monthly active"
+MONTH = 28
+# Recency in days of data which should be re-counted.
+INSTABLE_WINDOW = 30
```
acmiyaguchi: Nit: Unstable is more common than instable

tests/test_maudau.py
```diff
@@ -0,0 +1,127 @@
...
+        "20170503,1,3,20170620",
+        "20170504,1,3,20170620",
+        ""
+        ])
+    expected = (
+        '20170504',
+        [{'dau': '1', 'mau': '3', 'day': '20170503',
+          'generated_on': '20170620'}]
+    )
+    try:
```
acmiyaguchi: You probably don't want try-catch blocks when testing -- everything should be expected to work. It could lead to tests that unexpectedly succeed, but probably not in this case.

tests/test_maudau.py
```diff
@@ -0,0 +1,127 @@
...
+    assert expected0 == actual0, str(actual0)
+
+    expected28 = (
+        '2017-01-03', '2017-02-01', '20170103', '20170210'
+    )
+    actual28 = M.generate_filter_parameters(DT.date(2017, 1, 31), 28)
+    assert expected28 == actual28
+
+
+def test_generate_counts():
```
acmiyaguchi: If these two test cases are exercising two different edge cases, it would be helpful to either comment on what each case is doing, or split them into more specifically-named functions.

tests/test_maudau.py
```diff
@@ -0,0 +1,127 @@
...
+from mozetl.maudau import maudau as M
+
+NARROW_SCHEMA = StructType([
+    StructField("client_id",             StringType(),  True),
+    StructField("submission_date_s3",    StringType(),  False),
+    StructField("subsession_start_date", StringType(), True)])
+
+generated = M.format_as_submission_date(M.today.isoformat())
+
+spark = (
```
acmiyaguchi: There is a shared spark fixture with the same name in `conftest.py`. You can use it in any test function or fixture by adding it as a function parameter.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+              default='telemetry-parquet',
+              help='Bucket of the input dataset')
+@click.option('--input_prefix',
+              default='main_summary/v4',
+              help='Prefix of the input dataset')
+def main(input_bucket, input_prefix):
+    s3client = boto3.client('s3', 'us-west-2')
+    transferer = S3Transfer(s3client)
+    last_rollup_basename = get_last_rollup(transferer)
+    if last_rollup_basename:
```
acmiyaguchi: What happens when running this script back to back? I'd like to see a test case where this function is run multiple times on the same date with existing data -- or a comment explaining the assumptions of state between runs of this job.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+def count_active_users(frame, end_date, days_back):
+    """
+    Tally active users according to externally-defined heuristics,
+    specifically client-reported subsession_start_date.
+
+    :frame DataFrame(): conforming to main_summary schema
+    :end_date DT.date(): the last day on which to count activity.
+    :days_back int: how far back to go; 0 for daily and 28 for monthly.
+
+    Note that return values are not stable over time.
```
acmiyaguchi: Because of ping submission latency?

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+
+    See https://bugzilla.mozilla.org/show_bug.cgi?id=1240849
```
acmiyaguchi: Nit: Might be worth adding to the module docstring at the top.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
+"""
+Calculate Firefox monthly- and daily-active users.
```
acmiyaguchi: Please add a link to the original notebook or bug number for documentation purposes.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+
+def format_spark_path(bucket, prefix):
+    return "s3://{}/{}".format(bucket, prefix)
+
+
+def get_rollup_s3_paths(basename):
+    return (STORAGE_BUCKET, os.path.join(STORAGE_SUB_DIR, basename))
+
+
+def generate_filter_parameters(end_date, days_back):
```
acmiyaguchi: This might look better inline with the parent function. I would avoid returning more than 2 values in a function (a status, value tuple is one case where I think this style is somewhat appropriate).

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+    except botocore.exceptions.ClientError as e:
+        # If the file wasn't there, that's ok. Otherwise, abort!
+        if e.response['Error']['Code'] != "404":
+            raise e
+        else:
+            msg = "Did not find an existing file at '{}'".format(basename)
+            logger.info(msg)
+        return None
+    else:
+        return basename
```
acmiyaguchi: Nit: The `else` branch is unnecessary

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+def generate_counts(frame, since, until=None):
+    cols = [frame.client_id.alias('clientId'),
+            'subsession_start_date', 'submission_date_s3']
+    narrow = frame.select(cols)
+    updates = []
+    generated = format_as_submission_date(today.isoformat())
+    start = DT.datetime.strptime(since, '%Y%m%d').date()
+    until = (DT.datetime.strptime(until, '%Y%m%d').date() if until
+             else today)
+    while start < until:
```
acmiyaguchi: Is this functionally equivalent to a sliding window over a day and a month, each with a stride of 1?

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+        day = format_as_submission_date(start.isoformat())
+        d = {'day': day, 'dau': dau, 'mau': mau, 'generated_on': generated}
+        updates.append(d)
+        start += DT.timedelta(1)
+    return updates
+
+
+def write_locally(results):
+    '''
+    :results [{'day': '%Y%m%d', 'dau': <int>,
```
acmiyaguchi: This schema was useful :+1:

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,214 @@
...
+
+STORAGE_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
+STORAGE_SUB_DIR = 'spenrose/maudau' if DEVELOPMENT else 'mreid/maudau'
+MAUDAU_ROLLUP_BASENAME = "engagement_ratio.csv"
+MAUDAU_SNAPSHOT_TEMPLATE = "engagement_ratio.{}.csv"
+DASHBOARD_BUCKET = "net-mozaws-prod-metrics-data"
+DASHBOARD_KEY = "firefox-dashboard/{}".format(MAUDAU_ROLLUP_BASENAME)
+
+today = DT.date.today()
+yesterday = today - DT.timedelta(1)
```
SamPenrose: "yesterday" is not used; I deleted it. "today" is used as a default overrideable by the calling script except when it is used to record the date a file is written (which in turn is tied to the state of main_summary on S3).  I do not seee a purpose for overriding that; do you?

mozetl/maudau/maudau.py
SamPenrose: The function is broken out only for the purpose of exposing what should be unit tested.

mozetl/maudau/maudau.py
SamPenrose: Yes.

mozetl/maudau/maudau.py
SamPenrose: That sounds right.

mozetl/maudau/maudau.py
SamPenrose: Hhmm. Note that there are two "output" bucket/prefix pairs: STORAGE_BUCKET/STORAGE_SUB_DIR and DASHBOARD_BUCKET/DASHBOARD_KEY. Shall we continue to rely on the DEVELOPMENT flag to exclude the latter on staging? Should I move that flag to an environment variable (and if so, what)?

mozetl/maudau/maudau.py
SamPenrose: How about "main_summary is updated daily; re-runs should be idempotent between updates."?

mozetl/maudau/maudau.py
mreid-moz: This ETL job is a bit unusual in that it does not target a specific day. It looks at the existing data and determines *all* the days that need to be re-computed and updates each one that needs to be updated. This is different from how most of our airflow jobs work, but I think we should land it with the current logic first, then figure out how to refactor it to be more friendly to backfill operations / incremental dataset generation in a subsequent PR.

tests/test_maudau.py
SamPenrose: It's not catch but "finally", to restore the working directory. Any exceptions will still be raised. The alternative is to write a custom ContextManager or a decorator. Do you think I should do one of those?

mozetl/maudau/maudau.py
mreid-moz: You could introduce a class to hold these parameters in a single return value

tests/test_maudau.py
acmiyaguchi: A few options -- the first is to create a test fixture that sets up the directory and cleans up appropriately after the test, the second is to just change the directory at the end in the normal function state.

You could also just use relative paths from the temporary directory and not worry about changing the working path.

mozetl/maudau/maudau.py
acmiyaguchi: The `topline_dashboard` job is similar, but has a much lighter workload since it's just serializing parquet into csv. In any case, I think this should be moved outside of the global scope.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,221 @@
...
+    first date that needs to be re-counted.
+    """
+    start_date = start_date or DT.date.today()
+    since = (start_date - DT.timedelta(UNSTABLE_WINDOW))
+    since = format_as_submission_date(since)
+    carryover = []
+    with open(basename) as f:
+        reader = csv.DictReader(f)
+        for row in reader:
+            if row['day'] >= since:
```
mreid-moz: There is one more check that needs to happen for a given row - we need to make sure that the difference between `row['day']` and `row['generated']` is at least `ACTIVITY_SUBMISSION_LAG` days. That is, we want to ensure that the row was computed after the data became stable.

If we have an entry like:
`day: 20170610, dau: 100, mau: 200, generated_on: 20170615`
We should still re-compute it, even though the `day` is more than ten days ago (since it was computed with 5 days of data latency rather than the desired 10 days).

I think it would be better to split this function up into stages:
- parse the rollup to return all rows (this function minus the `break` and `since` pieces)
- filter rows to remove any "unstable" results, those being cases where `day` is recent or (`generated_on` - `day` is too low)
- Identify any resulting gaps. The previous code used `UNSTABLE_WINDOW` as an upper bound to how far back to look for such gaps, so you could do that as well, or simply process all identified gaps.

Then, we can pass in the specific set of days to re-compute to the `generate_counts` function.

These split-up functions should be fairly easy to test as well.

mozetl/maudau/maudau.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def generate_counts(frame, since, until=None):
+    cols = [frame.client_id.alias('clientId'),
+            'subsession_start_date', 'submission_date_s3']
+    narrow = frame.select(cols)
+    updates = []
+    today = DT.date.today()
+    generated = format_as_submission_date(today)
+    start = DT.datetime.strptime(since, '%Y%m%d').date()
```
mreid-moz: nit: consider introducing a `parse_as_submission_date` function similar to `format_as_submission_date`

tests/test_maudau.py
```diff
@@ -0,0 +1,117 @@
...
+    counts2 = M.generate_counts(frame, since, '20170504')
+    expected2 = {'day': '20170503', 'dau': 1, 'mau': 3,
+                 'generated_on': generated}
+    assert counts2[-1] == expected2, str(counts2[-1])
+
+
+def test_parse_last_rollup():
+    tempdir = tempfile.mkdtemp()
+    cwd = os.getcwd()
+    basename = "engagement_ratio.csv".format(generated)
```
mreid-moz: The `.format(..)` here has no effect without a placeholder in the format string

mozetl/maudau/maudau.py
SamPenrose: I think your logic is impeccable on its own terms, and I will code it if you want. That said:

1) The potential hole is an artifact of this wacky rolling-CSV-as-database approach.
2) There are no live examples of the potential hole in the existing CSVs I have checked (20170614, 20170114).
3) For a live example to be created, the existing job would have to not run for several weeks.
4) The proposal increases complexity and creates a code path that is very unlikely to run in production.

Do you still want me to do it?

mreid-moz: Hm, at the very least, we should avoid re-computing all 30 recent days every day and skip anything that accounts for the `ACTIVITY_SUBMISSION_LAG` period. The extra 20 days of processing would add a non-trivial amount of run time (if I recall correctly).

I agree that using a CSV file as the source of truth here is... suboptimal to say the least. We should address that general problem in a follow-up though, as it will probably require significant refactoring and testing to make it both idempotent and incremental.

SamPenrose: I replaced the 30-day window with the 10-day lag per request and also added a check for gaps and a test for the check. More date parameters have been moved from strings to date() instances.

mreid-moz:
There's a Travis CI failure that appears to be unrelated to these changes, but can you see if you can fix it?

SamPenrose:
Sure. Fix in this PR or separate?

mreid-moz:
This PR please, I don't like to merge while the build is failing.

SamPenrose:
The only reason I can see for this test to fail is that there is some path issue on Travis. The code is very straightforward and correct at a first scan. I pushed a tweak that gives Travis what it expects (and breaks the test on my machine) -- let's see if Travis is happy. Not sure we want to merge without understanding what is going wrong.

SamPenrose:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1376348 to follow up.

Issue #45, opened 2017-05-26
Add {posarg} to tox so we can run one test at a time


harterrt:
http://tox.readthedocs.io/en/latest/example/general.html

Issue #46, opened 2017-05-30
Package tests in a docker container


acmiyaguchi:
The current testing environment requires setting up `SPARK_HOME` and `PYTHONPATH`, then installing the proper dependencies before being able to run tests. `python_moztelemetry` uses docker to encapsulate the testing environment, and is easy to set up.

https://github.com/mozilla/python_moztelemetry/blob/master/Dockerfile
https://github.com/mozilla/python_moztelemetry/blob/master/runtests.sh

This project might benefit from setting up the environment in a similar fashion.

PR #57, opened 2017-06-29
Bug 1376808 - Package python_mozetl test infrastructure in docker

acmiyaguchi:
[Bug 1376808 - Package python_mozetl test infrastructure in docker](https://bugzilla.mozilla.org/show_bug.cgi?id=1376808)

Resolves issue #45 and #46. This should simplify the setup for running tests in `mozetl`. 

Travis might fail because flake8 is actually running now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,21 @@
-language: python
-python:
-    - "2.7"
+env:
+  global:
+    - CACHE_DIR=$HOME/.cache/docker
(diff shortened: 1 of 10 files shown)


harterrt:
Looks like tests are failing.

acmiyaguchi:
Tests are failing for flake8, those can be fixed in a separate PR.

PR #65, opened 2017-07-06
Bug 1378989 - remove catching of exceptions in churn job

acmiyaguchi:
[Bug 1378989 - Churn job from mozetl should retry on exceptions](https://bugzilla.mozilla.org/show_bug.cgi?id=1378989)

This should prevent the exception handler from suppressing the error. This behavior caused the churn job to miss data for several weeks.


mreid-moz:
Please fix flake8 complaint, then looks good to me!

PR #66, opened 2017-07-10
Bug 1368089 - Filter incontent searches in search aggregation jobs

acmiyaguchi:
In-content searches were removed from the search rollup in a refactor, causing
search data to become skewed. This also affects the topline summary, inflating
the search results significantly.

### DIFF
--- a/mozetl/constants.py
+++ b/mozetl/constants.py
@@ -1,0 +1,7 @@
+# Restrict to a whitelist of search_source's to avoid double counting while
+# we test our new search_count telemetry developed in:
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1367554
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1368089
+SEARCH_SOURCE_WHITELIST = [
+    'searchbar', 'urlbar', 'abouthome', 'newtab', 'contextmenu', 'system'
+]
--- a/mozetl/search/search_rollups.py
+++ b/mozetl/search/search_rollups.py
@@ -16,6 +16,7 @@ import click
 from pyspark.sql import SparkSession, functions as F
 
 from mozetl import utils
+from mozetl.constants import SEARCH_SOURCE_WHITELIST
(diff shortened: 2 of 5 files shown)


mozetl/search/search_rollups.py
```diff
@@ -45,6 +55,9 @@ def explode_searches(dataframe):
         dataframe
         .withColumn("search_counts", F.explode("search_counts"))
         .select(columns + exploded_search_columns)
+        .where("search_count > -1 or search_count is null")
```
mreid-moz: Please update the comment above to note that we're both exploding search counts and filtering out records from sources outside the above whitelist.

mozetl/topline/topline_summary.py
```diff
@@ -135,10 +135,16 @@ def search_aggregates(dataframe, attributes):
         .alias("count")
     )
 
+    source_whitelist = [
```
mreid-moz: Can we define the whitelist values in a single location shared by both topline and search rollups?

mozetl/constants.py
```diff
@@ -0,0 +1,7 @@
+# Restrict to a whitelist of search_source's to avoid double counting while
+# we test our new search_count telemetry developed in:
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1367554
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1368089
+source_whitelist = [
```
mreid-moz: nit: [recommended PEP8 style](https://www.python.org/dev/peps/pep-0008/#id48) is to use uppercase for constants. Also, if this file is going to be for general-purpose "constants", we should call this var something like `search_source_whitelist` for clarity.

acmiyaguchi:
[This notebook](https://gist.github.com/acmiyaguchi/addc6d93fcca5bf3aef8023d762db903) demonstrates that there is significant overcounting. This patch should bring values back down to normal.

Issue #67, opened 2017-07-12
Engagement Ratio / maudau export must apply special ACL headers


mreid-moz:
The [upload of data to the dashboard](https://github.com/mozilla/python_mozetl/blob/master/mozetl/maudau/maudau.py#L194) needs a special ACL header similar to the [topline dashboard export](https://github.com/mozilla/python_mozetl/blob/master/mozetl/utils.py#L52).

I think we should factor out an `upload_file_to_s3` function in `utils.py` that can be used by the existing `write_csv_to_s3` function as well as by the maudau job.

SamPenrose:
The file is updating correctly now.

PR #68, opened 2017-07-12
apply ACL headers to S3 transfer

SamPenrose:
Per issue #67. Spinning up a cluster to test.


mreid-moz:
Per IRC conversation, the test went fine.

PR #69, opened 2017-07-18
Bug 1381460 - Whitelist activity stream event source

piatra:
Should this be enough to track the new search event?

This pull request should not land until [a matching patch](https://bugzilla.mozilla.org/show_bug.cgi?id=1381460) on the Firefox side to add the event in telemetry gets r+. In order to make sure the event name won't change.


piatra:
@acmiyaguchi I can't assign a reviewer, can you please take this?

acmiyaguchi:
Sounds like this was successfully r+ into central, I'll merge this in.

PR #71, opened 2017-07-21
Create a dataset providing one flat row per (client_id, subsession_st…

SamPenrose:
Create a dataset providing one flat row per (client_id, subsession_start_date) in main_summary.

C.f. bug 1381896

### DIFF
--- a/mozetl/clientsdaily/fields.py
+++ b/mozetl/clientsdaily/fields.py
@@ -1,0 +1,177 @@
+from collections import defaultdict
+import pyspark.sql.functions as F
+
+VERSION = 1
+
+
+def mode(l, empty_value='MISSING'):
+    if not l:
+        return empty_value
+    counts = defaultdict(int)
+    for value in l:
+        counts[value] += 1
+    counts = [(v, k) for (k, v) in counts.items()]
+    counts.sort()
+    return counts[-1][1]
+
+
+MAIN_SUMMARY_FIELD_AGGREGATORS = [
+    F.sum('aborts_content').alias('aborts_content_sum'),
+    F.sum('aborts_gmplugin').alias('aborts_gmplugin_sum'),
+    F.sum('aborts_plugin').alias('aborts_plugin_sum'),
+    # active_addons
+    F.mean('active_addons_count').alias('active_addons_count_mean'),
+    F.first('active_experiment_branch').alias('active_experiment_branch'),
+    F.first('active_experiment_id').alias('active_experiment_id'),
+    # active_theme
+    F.sum(F.expr('active_ticks/(3600.0/5)')).alias('active_hours_sum'),
+    F.first('addon_compatibility_check_enabled').alias(
+        'addon_compatibility_check_enabled'),
+    F.first('app_build_id').alias('app_build_id'),
+    F.first('app_display_version').alias('app_display_version'),
+    F.first('app_name').alias('app_name'),
+    F.first('app_version').alias('app_version'),
+    # attribution
+    F.first('blocklist_enabled').alias('blocklist_enabled'),
+    F.first('channel').alias('channel'),
+    F.first('city').alias('city'),
+    F.first('country').alias('country'),
+    F.sum('crashes_detected_content').alias('crashes_detected_content_sum'),
+    F.sum('crashes_detected_gmplugin').alias('crashes_detected_gmplugin_sum'),
+    F.sum('crashes_detected_plugin').alias('crashes_detected_plugin_sum'),
+    F.sum('crash_submit_attempt_content').alias(
+        'crash_submit_attempt_content_sum'),
+    F.sum('crash_submit_attempt_main').alias('crash_submit_attempt_main_sum'),
+    F.sum('crash_submit_attempt_plugin').alias(
+        'crash_submit_attempt_plugin_sum'),
+    F.sum('crash_submit_success_content').alias(
+        'crash_submit_success_content_sum'),
+    F.sum('crash_submit_success_main').alias('crash_submit_success_main_sum'),
+    F.sum('crash_submit_success_plugin').alias(
+        'crash_submit_success_plugin_sum'),
+    F.first('default_search_engine').alias('default_search_engine'),
+    F.first('default_search_engine_data_load_path').alias(
+        'default_search_engine_data_load_path'),
+    F.first('default_search_engine_data_name').alias(
+        'default_search_engine_data_name'),
+    F.first('default_search_engine_data_origin').alias(
+        'default_search_engine_data_origin'),
+    F.first('default_search_engine_data_submission_url').alias(
+        'default_search_engine_data_submission_url'),
+    F.sum('devtools_toolbox_opened_count').alias(
+        'devtools_toolbox_opened_count_sum'),
+    F.first('distribution_id').alias('distribution_id'),
+    # userprefs/dom_ipc_process_count
+    F.first('e10s_cohort').alias('e10s_cohort'),
+    F.first('e10s_enabled').alias('e10s_enabled'),
+    F.first('env_build_arch').alias('env_build_arch'),
+    F.first('env_build_id').alias('env_build_id'),
+    F.first('env_build_version').alias('env_build_version'),
+    # events
+    # experiments
+    F.mean('first_paint').alias('first_paint_mean'),
+    # F.first(
+    #     F.expr("userprefs.extensions_allow_non_mpc_extensions"
+    #        ).alias("extensions_allow_non_mpc_extensions")
+    # ),
+    F.first('flash_version').alias('flash_version'),
+    F.first('install_year').alias('install_year'),
+    F.first('is_default_browser').alias('is_default_browser'),
+    F.first('is_wow64').alias('is_wow64'),
+    F.first('locale').alias('locale'),
+    # main
+    F.first('memory_mb').alias('memory_mb'),  # mean?
+    F.first('os').alias('os'),
+    F.first('os_service_pack_major').alias('os_service_pack_major'),
+    F.first('os_service_pack_minor').alias('os_service_pack_minor'),
+    F.first('os_version').alias('os_version'),
+    F.first('normalized_channel').alias('normalized_channel'),
+    F.countDistinct('document_id').alias('pings_aggregated_by_this_row'),
+    F.mean('places_bookmarks_count').alias('places_bookmarks_count_mean'),
+    F.mean('places_pages_count').alias('places_pages_count_mean'),
+    F.sum('plugin_hangs').alias('plugin_hangs_sum'),
+    F.sum('plugins_infobar_allow').alias('plugins_infobar_allow_sum'),
+    F.sum('plugins_infobar_block').alias('plugins_infobar_block_sum'),
+    F.sum('plugins_infobar_shown').alias('plugins_infobar_shown_sum'),
+    F.sum('plugins_notification_shown').alias(
+        'plugins_notification_shown_sum'),
+    # plugins_notification_user_action
+    # popup_notification_stats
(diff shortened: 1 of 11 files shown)


mozetl/clientsdaily/rollup.py
```diff
@@ -0,0 +1,121 @@
...
+from mozetl.utils import generate_filter_parameters, format_spark_path
+from fields import MAIN_SUMMARY_FIELD_AGGREGATORS
+
+ACTIVITY_SUBMISSION_LAG = DT.timedelta(10)
+MAIN_SUMMARY_VERSION = 4
+MAIN_SUMMARY_PATH = "s3://telemetry-parquet/main_summary/v{}".format(
+    MAIN_SUMMARY_VERSION)
+WRITE_VERSION = '2'
+STORAGE_BUCKET = 'net-mozaws-prod-us-west-2-pipeline-analysis'
+STORAGE_SUB_DIR = '/spenrose/clients-daily/v{}/'.format(WRITE_VERSION)
```
acmiyaguchi: I've seen the term `prefix` used more widely than `sub_dir` for s3 operations (`boto` for example).

mozetl/clientsdaily/rollup.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def load_main_summary(spark):
+    return spark.read.option(
```
acmiyaguchi: Formatting nit: I prefer using parenthesis to create a new scope where the method operator can work on new lines. For examples:

```
return (
    spark
    .read
    .option("mergeSchema", "true")
    .parquet(MAIN_SUMMARY_PATH)
)
```

If you are vehemently opposed to parentheses, I would suggest indenting the parenthesis after `"mergeSchema", "true"` to the next line.

mozetl/clientsdaily/rollup.py
```diff
@@ -0,0 +1,121 @@
...
+        "mergeSchema", "true").parquet(MAIN_SUMMARY_PATH)
+
+
+def extract_search_counts(frame):
+    '''
```
acmiyaguchi: Please use triple double quotes for docstrings

mozetl/clientsdaily/rollup.py
```diff
@@ -0,0 +1,121 @@
...
+    exploded = two_columns.select(
+        "did", F.explode("search_counts").alias("search_struct"))
+    unpacked = exploded.select(
+        "did",
+        F.expr("search_struct.count").alias("search_count_atom")
+    )
+    grouped = unpacked.groupBy("did").agg({"search_count_atom": "sum"})
+    extracted = grouped.select(
+        "did", F.col("sum(search_count_atom)").alias("search_count"))
+    nulls = two_columns.select(
```
acmiyaguchi: Please reformat this again, either nested in a new context or with the closing parenthesis on a new line.

mozetl/clientsdaily/rollup.py
```diff
@@ -0,0 +1,121 @@
...
+    result = frame.join(intermediate, frame.document_id == intermediate.did)
+    return result
+
+
+def extract_month(first_day, frame):
+    '''
+    :first_day DT.date(Y, m, 1)
+    :frame DataFrame homologous with main_summary
+    '''
+    day = first_day
```
acmiyaguchi: Choice of variable name is initially confusing, in particular `day.month`.

mozetl/clientsdaily/rollup.py
```diff
@@ -0,0 +1,121 @@
...
+        "search_counts is NULL").select(
+        "did", F.lit(0).alias("search_count")
+    )
+    intermediate = extracted.unionAll(nulls)
+    result = frame.join(intermediate, frame.document_id == intermediate.did)
+    return result
+
+
+def extract_month(first_day, frame):
+    '''
```
acmiyaguchi: One sentence description of what `extract_month` is doing.

mozetl/schemas.py
```diff
@@ -0,0 +1,10 @@
+import ujson as json
+import os
+
+from pyspark.sql.types import StructType
+
+directory = os.path.join(os.path.dirname(__file__), 'json')
```
acmiyaguchi: This might not work when packaged as a binary. It might be worthwhile to pull the logic from [`topline.schema.schema_from_json`](https://github.com/mozilla/python_mozetl/blob/master/mozetl/topline/schema.py) into the utility library and use it here. It uses `pkg_resource` to ensure that it references the json file correctly.

tests/test_clientsdaily.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.fixture
+def make_frame(spark):
+    root = os.path.dirname(__file__)
+    path = os.path.join(root, 'resources',
+                        'main_summary-late-may-1123-rows-anonymized.json')
+    with open(path) as f:
+        jrows = json.load(f)
+        jrdd = spark.sparkContext.parallelize(map(json.dumps, jrows))
+        frame = spark.read.json(jrdd, MAIN_SUMMARY_SCHEMA)
```
acmiyaguchi: Can you read the json directly?

tests/test_clientsdaily.py
```diff
@@ -0,0 +1,69 @@
...
+    return frame
+
+
+@pytest.fixture
+def make_frame_with_extracts(spark):
+    from mozetl.clientsdaily import rollup as R
```
acmiyaguchi: Imports should be module level. I would also avoid the 1 letter aliasing of namespaces (`pyspark.sql.functions as F` is contradictory to this statement, but it's also used as part of a domain specific language that has a fair number of conflicts with python built-ins).

mozetl/clientsdaily/fields.py
```diff
@@ -0,0 +1,177 @@
...
+    #     F.expr("userprefs.extensions_allow_non_mpc_extensions"
+    #        ).alias("extensions_allow_non_mpc_extensions")
+    # ),
+    F.first('flash_version').alias('flash_version'),
+    F.first('install_year').alias('install_year'),
+    F.first('is_default_browser').alias('is_default_browser'),
+    F.first('is_wow64').alias('is_wow64'),
+    F.first('locale').alias('locale'),
+    # main
+    F.first('memory_mb').alias('memory_mb'),  # mean?
```
SamPenrose: Fieldwise decisions are currently managed here: https://docs.google.com/spreadsheets/d/1jDqnhXrix8WtfBMrq1NzesYd7HrVn__VsA8RyckUrSo/edit#gid=868109559

mozetl/clientsdaily/rollup.py
SamPenrose: 1) +1 on conforming to the team style. 2) I'd like to conform to the *team* style. So no opposition from me, vehement or otherwise, but can we work this out on a shared policy level rather than a per-PR level?

fbertsch:
Drive-by comment, @SamPenrose and @acmiyaguchi. I'm really looking forward to this nice "here is the recommended way to aggregate these values" way of making this dataset (e.g. mean, max, sum, etc.). Might I recommend taking that document and moving it to an easy-to-read file that is part of this PR, which the code then reads to create the dataset? Then there won't ever be a separation between docs and code.

Maybe: http://docs.astropy.org/en/stable/io/ascii/

SamPenrose:
@fbertsch 1) Yes. 2) Not yet. 3) The code is ours; the policy should be driven by the analysts (what is valid / useful) informed by us (what is computationally expensive, has non-obvious footguns, etc).

fbertsch:
@SamPenrose I may not have been clear - regardless of who owns the code/definitions, this will make it easier to:

1. Comment on and change existing definitions
2. Use these same definitions in other places - for instance, if we wanted to add per-user aggregations to the experiment pipeline, I could easily see us using the definition file and generating the same measurements to compare across branches.

mozetl/schemas.py
SamPenrose: Thanks! Turns out pkg_resources.resource_stream() is a ContextManager which returns a file object. Would you like me to patch topline to use a with ... : block or should I leave it alone?

tests/test_clientsdaily.py
SamPenrose: Not sure what you mean.

tests/test_clientsdaily.py
SamPenrose: module-level import reveals a race condition in Pyspark initialization that ends up with "AttributeError: 'NoneType' object has no attribute '_jvm'"  when trying to invoke pyspark.sql.functions.sum(). If you have a clean solution for the race condition awesome otherwise I'd rather have this modest funkiness in the test module. What do you think?

mozetl/clientsdaily/fields.py
```diff
@@ -0,0 +1,177 @@
...
+        return empty_value
+    counts = defaultdict(int)
+    for value in l:
+        counts[value] += 1
+    counts = [(v, k) for (k, v) in counts.items()]
+    counts.sort()
+    return counts[-1][1]
+
+
+MAIN_SUMMARY_FIELD_AGGREGATORS = [
```
SamPenrose: I'd like to keep this reasonably declarative. Do you have a specific suggestion?

mozetl/clientsdaily/rollup.py
acmiyaguchi: +1, I might have worded the comment a bit strongly.  I think code style quickly becomes a bike shedding exercise, but PEP8 is a useful tool here. There is also a nice visual appeal for clauses to be on their own line, like `select`'s and `where`'s.

mozetl/schemas.py
acmiyaguchi: A `with` block sounds good.

tests/test_clientsdaily.py
acmiyaguchi: I'm wondering if `read.json` accepts the json file directly ([this example](https://spark.apache.org/docs/1.6.1/sql-programming-guide.html#creating-dataframes) reads from json, but doesn't show the shape of the data). Not an issue blocking this review.

tests/test_clientsdaily.py
acmiyaguchi: That's a pretty odd exception. Does this happen when you import `extract_search_counts`, `to_profile_day_aggregates`, and `extract_month` directly?

SamPenrose: Importing specific names does not bypass the problem. Cyclic import race conditions used to be A Thing in Python; note that the problem here is at the Python-as-JVM-child level. I should probably file a PySpark bug.

acmiyaguchi:
Sounds like a good idea. The `agg` operator takes a dictionary that [maps strings to strings](http://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.GroupedData.agg), the non-trivial part would be defining custom aggregators (like [`active_ticks`](https://github.com/mozilla/python_mozetl/pull/71/files#diff-6bcf55ce5b51faf06097e26d95bbe78bR27)).

tests/test_clientsdaily.py
SamPenrose: Right you are, thanks for the catch.

SamPenrose:
Latest update adds write_by_activity_day() and forces shell caller to specify the month.

SamPenrose:
OK, I think we're down to two or three issues: the local imports in the test module, your opening comment about two commits, and maybe concerns about fields. When you have a chance would you mind summarizing where you are on this PR?

acmiyaguchi:
The local imports bothers me, but if you the issue is filed somewhere, all is well. Multiple commits are not necessary by all means, but adding the bug number to the commit message would be good. Finally, if the fields are reflective of the document, I'm all for it. It might be good to file a follow up regarding @fbertsch's suggestion.

acmiyaguchi:
Looks good!

PR #73, opened 2017-07-24
DO-NOT-MERGE: start shield ETL job for privacy prefs study

groovecoder:
Need an initial review of my approach here. I tried converting the old blok notebook into an ETL job here.

@harterrt - If this looks like the good/right approach, I'll expand the `DATAFRAME_COLUMN_CONFIGS` to capture the rest of the ping data.

Metrics doc is here:
https://github.com/mozilla/shield-study-privacy/blob/master/docs/metrics.md

### DIFF
--- a/mozetl/shield/privacy_prefs.py
+++ b/mozetl/shield/privacy_prefs.py
@@ -1,0 +1,40 @@
+# coding: utf-8
+
+# Started from mashing up:
+#  https://gist.githubusercontent.com/ilanasegall/b3ce1aa0d3cc8c117a35b4a4fb9d4681/raw/c8a96e823cd56072e896e4c2d94c496306b59c8c/blok_df.py
+# with:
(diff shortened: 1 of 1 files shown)


mozetl/shield/privacy_prefs.py
```diff
@@ -0,0 +1,100 @@
...
+        sqlContext,
+        pings,
+        DataFrameConfig(DATAFRAME_COLUMN_CONFIGS, include_shield_pings)
+    )
+
+
+def include_shield_pings(ping):
+    return ping['payload/test'] == SHIELD_ADDON_ID
+
+shield_etl_job = testpilot_etl_boilerplate(
```
harterrt: Nit: You could just name this variable `etl_job` and avoid the second function below. We have a separate function definition in the containers job because we need to run two ETL jobs at once.

PR #76, opened 2017-07-25
mozetl/shield/privacy_prefs.py

groovecoder:
This reverts commit 4083764b4c62be941c8dec9df0e6c7af03eed416 to restore the `privacy_prefs.py` job; only now it can get an actual review instead of just pushed straight to `master`! 😢 

@harterrt - any way you can run this to check for errors and see if it catches the few `@shield-study-privacy` pings I've been sending while testing?

### DIFF
--- a/mozetl/shield/privacy_prefs.py
+++ b/mozetl/shield/privacy_prefs.py
@@ -17,7 +17,8 @@ DATAFRAME_COLUMN_CONFIGS = [
     ("event", "payload/payload/event", None, StringType()),
     ("originDomain", "payload/payload/originDomain", None, StringType()),
     ("breakage", "payload/payload/breakage", None, StringType()),
-    ("notes", "payload/payload/notes", None, StringType())
+    ("notes", "payload/payload/notes", None, StringType()),
+    ("study", "payload/payload/study", None, StringType()),
 ]
 
 
--- a/mozetl/shield/privacy_prefs.py
+++ b/mozetl/shield/privacy_prefs.py
@@ -8,7 +8,7 @@
 from pyspark.sql.types import StringType
 
 from ..basic import convert_pings, DataFrameConfig
-from .utils import testpilot_etl_boilerplate
+from ..testpilot.utils import testpilot_etl_boilerplate
 
 
 SHIELD_ADDON_ID = '@shield-study-privacy'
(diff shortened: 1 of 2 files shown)


groovecoder:
I added some tests, but I don't understand why they're failing. 😢 I thought I put the data in the tests in all the right format/places.

mozetl/shield/privacy_prefs.py
```diff
@@ -0,0 +1,40 @@
...
+def transform_shield_pings(sqlContext, pings):
+    return convert_pings(
+        sqlContext,
+        pings,
+        DataFrameConfig(DATAFRAME_COLUMN_CONFIGS, include_shield_pings)
+    )
+
+
+def include_shield_pings(ping):
+    return ping['payload/payload/study'] == SHIELD_ADDON_ID
```
harterrt: Looks like this key does not exist in the example pings. Should this be `payload/study`?

groovecoder: That's what I thought, but it's in both `payload/study` and `payload/payload/study` in [the test/example pings](https://github.com/mozilla/python_mozetl/pull/76/files#diff-b8e880181172dc46e07525f4610b926fR24) ...
```
input_payload = {
        'study': '@shield-study-privacy',
        'branch': 'thirdPartyCookiesOnlyFromVisited',
        'originDomain': 'www.paypal.com',
        'event': 'page-problem',
        'breakage': '',
        'notes': '',
        'study_version': '0.0.1',
        'about': {
            '_src': 'addon',
            '_v': 2
        }
    }
```

tests/test_shield_privacy_prefs.py
```diff
@@ -0,0 +1,167 @@
...
+        {'payload': {
+            'study': '@shield-study-privacy',
+            'other-ignored-field': 'who cares',
+            'payload': payload
+        }}
+    ])
+
+
+def create_row(overrides):
+    keys = ["branch", "event", "originDomain", "breakage", "notes"]
```
harterrt: Is the `study` key being dropped because it's not included in this array?

groovecoder: I don't think so - I tried including it here and got the same error.

groovecoder:
@harterrt - need another r? here after I merged your change and finished fixing up the tests.

groovecoder:
Rebased and merged. When can I expect to be able to write STMO queries with `SELECT * FROM privacy_prefs_shield` ?

harterrt:
We'll need to schedule this job on ATMO. Once it's running we'll need to open a bug to get it added to Presto, as [described here](https://docs.telemetry.mozilla.org/cookbooks/create_a_dataset.html). I can set this up later this week. Does that work?

groovecoder:
ASAP is best, but yeah. Any chance to run it manually a couple times just to populate some test data so I can start my SQL queries & dashboard?

Issue #38, opened 2017-05-22
Standardize CLI entrypoint for airflow jobs


harterrt:
We have a few different access points configured for different jobs [1]. We should have a single entrypoint that reads common environment variables, starts a spark context, and launches the proper job. 

[1] Examples:
https://github.com/mozilla/python_mozetl/blob/master/scheduling/pulse.py
https://github.com/mozilla/python_mozetl/blob/master/scheduling/tab_spinner.py
https://github.com/mozilla/python_mozetl/blob/master/mozetl/churn/churn.py#L648

acmiyaguchi:
I think this take the following form:

* Using options and required options over positional arguments
* Add job to an `entry_point` object
* A common entry point in `telemetry-airflow/jobs`

acmiyaguchi:
There are some subtle side-effects of using automatic environment variables that should be documented.

There is an option called `auto_envvar_prefix` in commands that wraps all `click.options` with the ability to read the value from the environment. This is done by taking the name of the prefix and appending the name of the option. In addition, if the command is part of a group (the mozetl entry point), then the name of the subcommand is appended to prefix. For example, the `date` option from the `churn` job would read the value `MOZETL_CHURN_DATE` from the environment if `auto_envvar_prefix="MOZETL"`.

A utility function could be added to aid in the construction of the environment variables in the airflow dag.

```
def mozetl_envvar(command, env):
    # TODO: also include command name as an environment variable, for mozetl job script
    return {"MOZETL_{}_{}".format(command, key): value for key, value in env}
```

acmiyaguchi:
The `mozetl-submit` can be added to the `setup.py` with custom build steps. It might also be useful for building the docker container and running tests.

acmiyaguchi:
This has been merged in, now all is left to do is to add more jobs to it.

PR #81, opened 2017-07-30
Bug 1385232 - Add unified entry point and submission script

acmiyaguchi:
Addresses issue #38. Updates churn to use the new cli entry point, and adds generalized airflow script that can be used locally and in production.


acmiyaguchi:
Yes, the date and bucket environment variables will need to change in the dag. The wrapper is already written, but I'm waiting for changes to settle and an opportunity to test this locally.

bin/mozetl-submit.sh
```diff
@@ -0,0 +1,88 @@
...
+    cd python_mozetl
+fi
+
+pip install .
+python setup.py bdist_egg
+
+spark-submit --master ${MOZETL_SPARK_MASTER} \
+             --deploy-mode client \
+             --py-files dist/*.egg \
+             ${workdir}/runner.py ${MOZETL_ARGS}
```
harterrt: Instead of passing a multiple parameters of the form `MOZETL_JOBNAME_VARIABLENAME` could we just pass one variable `MOZETL_COMMAND` of the form `"job_name --option1 value --option2 value2"`? 

The advantage would be that we don't need to populate any click arguements from env variables. It might also be easier to read in airflow.

acmiyaguchi: I think in this is setup, you could probably set `MOZETL_COMMAND` to the arguments and have it work as expected. The `mozetl_envvar` wrapper lets us specify the options as a dictionary without changing the current workflow too much.

PR #87, opened 2017-08-02
bug 1386446: fix shield doc type

groovecoder:


groovecoder:
Happy to try ... I'm launching a cluster now, though I've forgotten the docs on how to run this particular job on it?

groovecoder:
I took a stab at converting this into a `shield-study-privacy-etl` notebook, but I'm not sure if/how I can populate the `sc` and `sqlContext` arguments to `etl_job` there?

https://gist.github.com/groovecoder/fbd520d8c9ab3e59fa50579b4d0dcf11

harterrt:
`sc` and `sqlContext` are provided by default in the jupyter notebook, i.e. they should already be defined.

groovecoder:
@harterrt - I'm having a hard time re-running my job in this notebook with `sc` and `sqlContext` passed into the `etl_job()` call. 😢 Any chance you can try to run it for me?

harterrt:
Sure, I can take a look tomorrow (Thursday).

groovecoder:
Note: see https://bugzilla.mozilla.org/show_bug.cgi?id=1387236 for updates on debugging the lower-level Telemetry/ping payload.

harterrt:
OK - this should work after https://github.com/mozilla/python_mozetl/pull/87/commits/e2c601311989e6dedce58733ff9b28cd12257df2. I backfilled over 7 days. Take a look at the data in re:dash and make sure everything looks right.

Thanks for working through this with us!

harterrt:
Assuming the tests pass, we can commit this in the morning. I've scheduled the job on ATMO to run at Noon daily using the following notebook:
https://gist.github.com/harterrt/113a15c527b2897314fed0d8bce74fab

harterrt:
@mreid-moz - I made some config changes to this PR, so I don't want to be the sole reviewer. Can you PTAL since you have some context from [Bug 1387236](https://bugzilla.mozilla.org/show_bug.cgi?id=1387236).

groovecoder:
I don't have any Re:dash queries set up yet. But I should be able to do some spot-check queries with `SELECT * FROM privacy_prefs_shield ...` ?

groovecoder:
After @harterrt's back-fill, I was able to see 7k+ records, including a bunch with all the breakage details & notes. r+ from me!!

groovecoder:
Hopefully the extra week of back-fill data (we launched the experiment on 2017-07-27) will show more clients ... right now I only see:
```
control | 29
firstPartyIsolation | 34
firstPartyIsolationOpenerAccess | 18
noThirdPartyCookies | 40
originOnlyRefererToThirdParties | 42
resistFingerprinting | 31
sessionOnlyThirdPartyCookies | 19
thirdPartyCookiesOnlyFromVisited | 21
trackingProtection | 30
```
Too few participants for significant data analysis. 😒 If the full back-fill doesn't show many more participants, we'll have to see if we can increase our recruitment efforts.

mozetl/shield/privacy_prefs.py
```diff
@@ -8,17 +8,18 @@
...
 
 SHIELD_ADDON_ID = '@shield-study-privacy'
 DATAFRAME_COLUMN_CONFIGS = [
-    ("branch", "payload/payload/branch", None, StringType()),
-    ("event", "payload/payload/event", None, StringType()),
-    ("originDomain", "payload/payload/originDomain", None, StringType()),
-    ("breakage", "payload/payload/breakage", None, StringType()),
-    ("notes", "payload/payload/notes", None, StringType()),
-    ("study", "payload/payload/study", None, StringType()),
+    ("client_id", "clientId", None, StringType()),
```
mreid-moz: I think a code comment is warranted here - referencing bug 1387236 and the non-standard field names

mozetl/shield/privacy_prefs.py
```diff
@@ -31,11 +32,11 @@ def transform_shield_pings(sqlContext, pings):
...
 def include_shield_pings(ping):
-    return ping['payload/payload/study'] == SHIELD_ADDON_ID
+    return ping['payload/study'] == SHIELD_ADDON_ID
 
 
 def etl_job(sc, sqlContext, **kwargs):
-    return testpilot_etl_boilerplate(
+    return shield_etl_boilerplate(
         transform_shield_pings,
         's3n://telemetry-parquet/harter/privacy_prefs_shield/v1'
```
mreid-moz: Aside: is this the correct path? Should the ETL code accept the target path via an env variable passed from Airflow?

mozetl/shield/privacy_prefs.py
```diff
@@ -31,11 +32,11 @@ def transform_shield_pings(sqlContext, pings):
...
-    )
+    )(sc, sqlContext)
```
mreid-moz: Should we pass the target `submission_date` through from Airflow to allow backfills?

harterrt: I don't expect to schedule this job on Airflow. These shield and testpilot ETL jobs tend to be short lived. ATMO does well enough.

mozetl/shield/privacy_prefs.py
harterrt: As noted below, I don't expect to schedule this on Airflow.

harterrt: Filed https://github.com/mozilla/python_mozetl/issues/97

mreid-moz:
@groovecoder note that the example submission shown in [bug 1387236](https://bugzilla.mozilla.org/show_bug.cgi?id=1387236#c9) does not contain a `clientId`, so you may need some other mechanism of counting enrolment (only 264 of the 7k records contain a value for `clientId`). 

This is the same number of records that contain a value for `branch` - the vast majority of records show no value for `branch`. Maybe there's something else going on with the data...

mozetl/shield/privacy_prefs.py
harterrt: Done

mozetl/shield/utils.py
```diff
@@ -0,0 +1,27 @@
...
+        ).where(
+            docType="shield-study",
+            submissionDate=submission_date,
+            appName="Firefox",
+        ).records(sc)
+
+        transformed_pings = transform_func(sqlContext, pings)
+
+        if save:
+            # path = 's3://telemetry-parquet/testpilot/txp_pulse/v1/submission_date={}'
```
harterrt: Done

harterrt:
Alright - I made one final change in https://github.com/mozilla/python_mozetl/pull/87/commits/5adb9de28c7867aa62eab5467d885b67d332381d. It looks like my backfill didn't work because `etl_job` did not respect the `submission_date` parameter, as Mark pointed out.

This also explains why there are so many null `client_id` fields. We only have data for 2017-08-08 because the backfill wasn't respecting `submission_date`. We also have some errant data for 2017-08-02 in the dataset from a past. [This table](https://sql.telemetry.mozilla.org/queries/15436/source) is illustrative.

Once this is approved, I'll restart the backfill to fix this issue.

groovecoder:
Ah, I wonder if the pings for `installed|startup|shutdown` have different formats than the others?

groovecoder:
🙌 🎉 Is the notebook running to back-fill the rest of the data? Do we have an ETA when it will be done? I still just see the `20170808` and `20170802` records in `privacy_prefs_shield`

harterrt:
After some false starts this morning, the backfill appears to be done: https://gist.github.com/harterrt/6d549dbf1720c21b52bf652a4069061c. Last I checked STMO is still catching up.

PR #88, opened 2017-08-02
Fix txp_mau_dau job for case where the experiment doesn't have an addon

sunahsuh:
The "Send" experiment doesn't use an addon at all, which broke an assumption in the txp_mau_dau job. Adds test for this case and the public testpilot `experiments.json` file as a fixture.

### DIFF
--- a/tests/test_txp_mau_dau.py
+++ b/tests/test_txp_mau_dau.py
@@ -1,0 +1,21 @@
+from mozetl.testpilot.txp_mau_dau import get_experiments
+from utils import get_resource
+import requests
+import json
+
+
+class MockRequest():
+    def raise_for_status(self):
+        pass
+
+    def json(self):
+        return json.loads(get_resource('txp_experiments.json'))
+
+
+def test_get_experiments(monkeypatch):
+    def mock_get(url):
+        return MockRequest()
+    monkeypatch.setattr(requests, 'get', mock_get)
+    # There are 14 total experiments in the file but one should be filtered out
+    # due to not having an addon_id -- we base this mau/dau on the addons dataset
+    assert len(get_experiments()) == 13
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,0 +1,20 @@
+import os
+
+
+def get_resource_path(filename):
+    """Returns the fully qualified path for a file in the test resources directory
+
+    :param filename: the file's name as a string
+    :return: the fully qualified path for the file as a string
+    """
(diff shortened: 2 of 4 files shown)


tests/test_txp_mau_dau.py
```diff
@@ -0,0 +1,21 @@
...
+    def json(self):
+        return json.loads(get_resource('txp_experiments.json'))
+
+
+def test_get_experiments(monkeypatch):
+    def mock_get(url):
+        return MockRequest()
+    monkeypatch.setattr(requests, 'get', mock_get)
+    # There are 14 total experiments in the file but one should be filtered out
+    # due to not having an addon_id
```
harterrt: Why do we want to ignore experiments without addons? Is it because we do not care about these experiments or because we can't identify active users without using an addon_id? Can you add a comment explaining why we don't care or why we're making this compromise?

tests/utils.py
```diff
@@ -0,0 +1,20 @@
+import os
+
+
+def get_resource_path(filename):
```
acmiyaguchi: I've had issues with static resource (`json` files for schema definitions) when packaged. I would encourage the use of [`pkg_resources`](http://setuptools.readthedocs.io/en/latest/pkg_resources.html)

tests/test_txp_mau_dau.py
sunahsuh: Updated -- but, given we're basing these mau/dau numbers off of the addons dataset, we don't really care about the non-addon experiments (and those are more accurately tracked on the serverside anyway since they're webapps)

tests/utils.py
sunahsuh: Oh, interesting -- do we have a usage example elsewhere in the repo? I went with this since this appears to be the current pattern in other tests, but it's repeated in a few places (the intention being that we'd DRY the other instances up over time.)

acmiyaguchi: You'd probably be okay with using this just in the tests since they're usually not included in the package. I've used pkg_resources when defining the [schemas for topline](https://github.com/mozilla/python_mozetl/blob/master/mozetl/topline/schema.py). It's also being used to define [the main_summary schema](https://github.com/mozilla/python_mozetl/blob/master/mozetl/schemas.py#L12).

sunahsuh: That's super neat, thanks for the pointer! I'll probably stick with this for now so I can fix this job asap

PR #89, opened 2017-08-03
Make privacy prefs filter pings properly

harterrt:
This makes the job recognize the shield pings. We still have the problem where fields are not showing up in the raw pings.


groovecoder:
Tests need to be updated to match.

PR #91, opened 2017-08-04
add several dimensions and write more efficiently

SamPenrose:
Work for bug 1381896:

1) Repartitions before writing Parquet files and skips _SUCCESS flag for redash compatibility.
2) Adds dimensions: profile_age_in_days, sessions_started_on_this_day.
3) Casts profile_creation_date to a date object.

### DIFF
--- a/mozetl/clientsdaily/fields.py
+++ b/mozetl/clientsdaily/fields.py
@@ -97,9 +97,15 @@ MAIN_SUMMARY_FIELD_AGGREGATORS = [
         'plugins_notification_shown_sum'),
     # plugins_notification_user_action
     # popup_notification_stats
-    F.first('profile_creation_date').alias('profile_creation_date'),
-    F.sum('push_api_notification_received').alias(
-        'push_api_notification_received_sum'),
+    F.first(
+        F.expr(
+            "datediff(subsession_start_date, " \
+            "from_unixtime(profile_creation_date*24*60*60))"
+        )
+    ).alias("profile_age_in_days"),
+    F.first(
+        F.expr("from_unixtime(profile_creation_date*24*60*60)")
+    ).alias('profile_creation_date'),
     F.sum('push_api_notify').alias('push_api_notify_sum'),
     F.first('sample_id').alias('sample_id'),
     F.first('scalar_parent_aushelper_websense_reg_version').alias(
@@ -157,6 +163,8 @@ MAIN_SUMMARY_FIELD_AGGREGATORS = [
     F.first('search_cohort').alias('search_cohort'),
     F.sum('search_count').alias('search_count_sum'),
     F.mean('session_restored').alias('session_restored_mean'),
+    F.sum(F.expr("IF(subsession_counter = 1, 1, 0)")).alias(
+        "sessions_started_on_this_day"),
     # shutdown_kill
     F.sum(F.expr('subsession_length/3600.0')).alias('subsession_hours_sum'),
     # ssl_handshake_result
--- a/mozetl/clientsdaily/rollup.py
+++ b/mozetl/clientsdaily/rollup.py
@@ -6,8 +6,7 @@ import click
 from moztelemetry.standards import filter_date_range
 from mozetl.utils import (
     format_spark_path,
-    generate_filter_parameters,
-    delete_from_s3
+    generate_filter_parameters
 )
 from fields import MAIN_SUMMARY_FIELD_AGGREGATORS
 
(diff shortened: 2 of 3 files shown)


SamPenrose:
The test appears to be failing due to either timezone or sorting issues. I need to drop a field that has been removed from main_summary, so update coming.

SamPenrose:
The call | from_unixtime(profile_creation_date*24*60*60) | is resulting in off-by-one errors, probably because of implicit TZ handling. I'll patch.

mozetl/clientsdaily/fields.py
```diff
@@ -97,9 +97,15 @@ def mode(l, empty_value='MISSING'):
         'plugins_notification_shown_sum'),
     # plugins_notification_user_action
     # popup_notification_stats
-    F.first('profile_creation_date').alias('profile_creation_date'),
-    F.sum('push_api_notification_received').alias(
-        'push_api_notification_received_sum'),
+    F.first(
+        F.expr(
+            "datediff(subsession_start_date, " \
```
acmiyaguchi: Nit: If you call the function directly instead of through `expr`, you can use a line break instead of using a line continuation.

mozetl/clientsdaily/fields.py
```diff
@@ -157,6 +163,8 @@ def mode(l, empty_value='MISSING'):
     F.first('search_cohort').alias('search_cohort'),
     F.sum('search_count').alias('search_count_sum'),
     F.mean('session_restored').alias('session_restored_mean'),
+    F.sum(F.expr("IF(subsession_counter = 1, 1, 0)")).alias(
+        "sessions_started_on_this_day"),
```
acmiyaguchi: This counts the number of first subsession profiles?

mozetl/clientsdaily/rollup.py
```diff
@@ -88,21 +87,31 @@ def to_profile_day_aggregates(frame_with_extracts):
...
         data_for_date = results.where(results.activity_date == isoday)
-        data_for_date.write.parquet(output_path)
-        # Remove file that prevents Parquet from rolling up.
-        keys_to_delete.append(os.path.join(prefix, '_SUCCESS'))
+        data_for_date.coalesce(partition_count).write.parquet(output_path)
         day_pointer += DT.timedelta(1)
-    delete_from_s3(output_bucket, keys_to_delete)
+
+
+def get_partition_count_for_writing(using_sample_id):
```
acmiyaguchi: Nit: `is_sampled` might be a nicer variable name.

mozetl/clientsdaily/rollup.py
```diff
@@ -137,7 +146,13 @@ def main(date, input_bucket, input_prefix, output_bucket,
         month_frame = month_frame.where(clause)
     with_searches = extract_search_counts(month_frame)
     results = to_profile_day_aggregates(with_searches)
-    write_by_activity_day(results, date, output_bucket, output_prefix)
+    partition_count = get_partition_count_for_writing(bool(sample_id))
+    spark.conf.set(
+        "mapreduce.fileoutputcommitter.marksuccessfuljobs", "false"
+    )  # Don't write _SUCCESS files, which interfere w/ReDash discovery
```
acmiyaguchi: Is the issue with `_SUCCESS` files documented else where? Is this a hive/presto or athena issue? If so, a bug number would be nice.

tests/test_clientsdaily.py
```diff
@@ -64,3 +64,61 @@ def test_to_profile_day_aggregates(spark):
...
+        u'2017-04-26', u'2015-06-18'
+    ])
+    expected_forward = set([
+        u'2014-12-18', u'2016-09-09',
+        u'2016-05-14', u'2017-02-18',
+        u'2012-11-19', u'2013-09-10',
+        u'2017-02-14', u'2016-04-06',
+        u'2017-04-27', u'2015-06-19'
+    ])
+    ten_pcds = clients_daily.select("profile_creation_date").take(10)
```
acmiyaguchi: I'm not entirely sure what's happening here. Does the ordering matter here, or is it just taking 10 random values? Does this test also verify that profile creation dates are valid DateTime types?

mozetl/clientsdaily/fields.py
SamPenrose: It counts the number of sessions started by this client on this day.

tests/test_clientsdaily.py
SamPenrose: 1. Per JIRA issue referenced in the comment a few lines up, Spark is sensitive to environment TZ but does not provide an API for controlling it. Therefore, any of three results can be correct.
2. Ordering doesn't matter.
3. The test doesn't validate that and, in fact, they aren't per the expression that creates actual1. Will fix in a followup PR.

PR #93, opened 2017-08-09
Add searchdb to unified entrypoint

harterrt:


harterrt:
Thanks, Anthony! Done

PR #100, opened 2017-08-14
bug 1387236: extract and load study_state too

groovecoder:
I want to measure the % of users who submit 1+ "page-problem" pings. But we only have data for the 'page-problem|page-works|breakage|notes|disable' pings. So, we don't know how many users are running the study who have never sent a page problem/works report.

This PR adds `study_state` extraction and load into the table, so I can get the total # of clients running the study add-on - including the ones who have not sent any breakage reports at all.

### DIFF
--- a/mozetl/shield/privacy_prefs.py
+++ b/mozetl/shield/privacy_prefs.py
@@ -5,40 +5,85 @@
 # with:
 # https://github.com/mozilla/python_mozetl/blob/689afa3d23229ca717422314c5a56abd83a85a0d/mozetl/testpilot/containers.py
 
+from datetime import date, timedelta
+
 from pyspark.sql.types import StringType
+from moztelemetry.dataset import Dataset
 
 from ..basic import convert_pings, DataFrameConfig
-from .utils import shield_etl_boilerplate
 
 
 SHIELD_ADDON_ID = '@shield-study-privacy'
 
 # Note that the JSON paths in this config are unusual for Shield pings.
-# See Bug 1387236 for more details and discussion
-DATAFRAME_COLUMN_CONFIGS = [
+# See https://bugzil.la/1387236 for more details and discussion
+
+COMMON_COLUMN_CONFIGS = [
     ("client_id", "clientId", None, StringType()),
     ("branch", "payload/branch", None, StringType()),
+    ("study_state", "payload/study_state", None, StringType()),
     ("event", "payload/event", None, StringType()),
     ("originDomain", "payload/originDomain", None, StringType()),
     ("breakage", "payload/breakage", None, StringType()),
     ("notes", "payload/notes", None, StringType()),
+]
+
+STUDY_STATE_DATAFRAME_COLUMN_CONFIGS = COMMON_COLUMN_CONFIGS + [
+    ("study", "payload/study_name", None, StringType()),
+]
+
+STUDY_EVENT_DATAFRAME_COLUMN_CONFIGS = COMMON_COLUMN_CONFIGS + [
     ("study", "payload/study", None, StringType()),
 ]
 
 
-def transform_shield_pings(sqlContext, pings):
+def transform_state_pings(sqlContext, pings):
     return convert_pings(
         sqlContext,
         pings,
-        DataFrameConfig(DATAFRAME_COLUMN_CONFIGS, include_shield_pings)
+        DataFrameConfig(STUDY_STATE_DATAFRAME_COLUMN_CONFIGS,
+                        include_state_pings)
     )
 
 
-def include_shield_pings(ping):
+def transform_event_pings(sqlContext, pings):
+    return convert_pings(
+        sqlContext,
+        pings,
+        DataFrameConfig(STUDY_EVENT_DATAFRAME_COLUMN_CONFIGS,
+                        include_event_pings)
+    )
+
+
+def include_event_pings(ping):
     return ping['payload/study'] == SHIELD_ADDON_ID
 
 
-etl_job = shield_etl_boilerplate(
-    transform_shield_pings,
-    's3n://telemetry-parquet/harter/privacy_prefs_shield/v1'
-)
+def include_state_pings(ping):
+    return ping['payload/study_name'] == SHIELD_ADDON_ID
+
+
+def etl_job(sc, sqlContext, submission_date=None, save=True):
+    s3_path = 's3n://telemetry-parquet/harter/privacy_prefs_shield/v1'
+    if submission_date is None:
+        submission_date = (date.today() - timedelta(1)).strftime("%Y%m%d")
+
+    pings = Dataset.from_source(
+        "telemetry"
+    ).where(
+        docType="shield-study",
+        submissionDate=submission_date,
+        appName="Firefox",
+    ).records(sc)
+
+    transformed_event_pings = transform_event_pings(sqlContext, pings)
+
+    transformed_state_pings = transform_state_pings(sqlContext, pings)
+
+    transformed_pings = transformed_event_pings.union(transformed_state_pings)
+
+    if save:
+        path = s3_path + '/submission_date={}'.format(submission_date)
+        transformed_pings.repartition(1).write.mode('overwrite').parquet(path)
+
+    return transformed_pings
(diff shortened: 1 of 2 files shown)


groovecoder:
Updated description of PR.

groovecoder:
Tests failed here, but probably/maybe for a good reason? What happens to this job when the `study_state` field isn't included? (Which happens on the in-study pings vs. the shield-state pings)

groovecoder:
@acmiyaguchi or @sunahsuh - can I get an r? here? In case @harterrt isn't able to review this week?

mozetl/shield/privacy_prefs.py
```diff
@@ -23,6 +23,7 @@
     ("breakage", "payload/breakage", None, StringType()),
     ("notes", "payload/notes", None, StringType()),
     ("study", "payload/study", None, StringType()),
+    ("study_state", "payload/study_state", None, StringType()),
```
acmiyaguchi: You should add `study_state` to the failing test inputs. From the travis logs:

```
>       assert row_to_dict(actual) == create_row(result_payload)
E       AssertionError: assert {'branch': 't...problem', ...} == {'branch': 'th...problem', ...}
E         Omitting 7 identical items, use -vv to show
E         Left contains more items:
E         {'study_state': None}
E         Use -v to get the full diff
```

If you omit the field, it still gets filled in as a null value because it's included with the table schema.

acmiyaguchi:
You should also formulate your question and assumptions about the inclusion of `study_state` as a test.

groovecoder:
Updated with more + better tests. I also introduce 2 separate transforms in the single ETL job: 1 to transform "state" pings and 1 to transform "event" pings. I believe I've written this correctly, but my tests for the new "state" pings are failing and I don't understand why. 😢 

### State Pings
These are sent by the underlying shield library. They use `study_name` instead of `study` and include `"study_state": "ineligible|install|running|user-ended-study"` values. E.g.,
```
{
  "study_name": "@shield-study-privacy",
  "branch": "firstPartyIsolationOpenerAccess",
  "study_state": "running",
  "study_version": "0.0.4",
  "about": {
    "_src": "shield",
    "_v": 2
  }
}
```

### Event pings
These are sent by *MY* study code. They use `study` and include `originDomain`, `event`, `breakage`, and `notes` fields & values. E.g.,
```
{
  "study": "@shield-study-privacy",
  "branch": "firstPartyIsolationOpenerAccess",
  "originDomain": "addons.mozilla.org",
  "event": "page-works",
  "breakage": "",
  "notes": "",
  "study_version": "0.0.4",
  "about": {
    "_src": "addon",
    "_v": 2
  }
}
```

tests/test_shield_privacy_prefs.py
```diff
@@ -12,16 +13,111 @@ def create_ping_rdd(sc, payload):
...
             "notes", "study"]
 
-    overrides['study'] = '@shield-study-privacy'
     overrides['client_id'] = 'aa'
 
     return {key: overrides.get(key, None) for key in keys}
 
 
+def create_state_row(overrides):
+    keys = ["client_id", "branch", "study", "study_state"]
```
acmiyaguchi: This should be `study_name` instead of `study`, as per your comment about state pings.

tests/test_shield_privacy_prefs.py
```diff
@@ -12,16 +13,111 @@ def create_ping_rdd(sc, payload):
...
+
+    overrides['client_id'] = 'aa'
+
+    return {key: overrides.get(key, None) for key in keys}
+
+
+def test_study_running(row_to_dict, spark_context):
```
acmiyaguchi: There doesn't seem to be a code path that checks whether the `study_state` is within a given whitelist of options. These tests seem pretty redundant.

mozetl/shield/utils.py
```diff
@@ -17,7 +17,7 @@ def etl_job(sc, sqlContext, submission_date=None, save=True):
 
         transformed_pings = transform_func(sqlContext, pings)
 
-        if save:
+        if type(save) == bool and save is True:
```
acmiyaguchi: What does does add?

tests/test_shield_privacy_prefs.py
groovecoder: When I did that, I got ...
```
>       assert row_to_dict(actual) == create_state_row(result_payload)
E       AssertionError: assert {'branch': 'f... 'ineligible'} == {'branch': 'fi... 'ineligible'}
E         Omitting 3 identical items, use -vv to show
E         Left contains more items:
E         {'study': u'@shield-study-privacy'}
E         Right contains more items:
E         {'study_name': '@shield-study-privacy'}
E         Use -v to get the full diff
```

That's why I'm confused. 😢

groovecoder: But I think this might be the right behavior, and I need to change the test to match? I.e., I **WANT** the transform to turn the state ping payload `study_name` into `study` in parquet table so it's consistent with the event pings in parquet.

mozetl/shield/utils.py
groovecoder: I had started to expand this functionality to accept save value of `True`, `overwrite`, or `____`? - i.e., whatever `write.mode()` does **NOT** overwrite existing records.

I'm worried that calling an `etl_job` twice from `privacy_prefs.py` with the same `s3_path` means the 2nd job will overwrite the first job's records?

acmiyaguchi: From what I understand, you don't want to have two separate datasets to process this dataset. If so, you should be taking the union of the two sets before writing them to disk.

On another note, the option you probably want is `mode="error"`. https://spark.apache.org/docs/latest/sql-programming-guide.html#save-modes

groovecoder:
Okay, this needs a new review because I changed it pretty substantially. Is this the right way to combine `transformed_pings` together before writing them to parquet?
```
transformed_pings = transformed_event_pings + transformed_state_pings
```
?

acmiyaguchi:
Not quite, you should probably use [`union` ie `tranformed_event_pings.union(transformed_state_pings)`](http://spark.apache.org/docs/2.1.0/api/python/pyspark.sql.html#pyspark.sql.DataFrame.union). You also have to reconcile the two dataframes because they now have very different schemas. I would definitely write a test that combines the result of those two datasets and runs a very simple sample query over them (similar to how you would use it in practice).

groovecoder:
Ready for next (last?) review ...

1. Updated to use `.union`
2. Consolidated `COMMON_COLUMN_CONFIGS` and used tiny `STUDY_STATE_DATAFRAME_COLUMN_CONFIGS|STUDY_EVENT_DATAFRAME_COLUMN_CONFIGS` to handle the different `study` paths
3. Added `test_combine_state_and_event` to test the union.

tests/test_shield_privacy_prefs.py
```diff
@@ -14,14 +15,105 @@ def create_ping_rdd(sc, payload):
...
+    # simulate transforuming study_name into study field
+    if 'study_name' in overrides:
+        del overrides['study_name']
+        overrides['study'] = '@shield-study-privacy'
     overrides['client_id'] = 'aa'
 
     return {key: overrides.get(key, None) for key in keys}
 
 
+def test_study_running(row_to_dict, spark_context):
```
acmiyaguchi: I still think these 4 tests are redundant. The only field that changes is `study_state`, but there is nothing that checks the validity of `study_state`. A better test would be an invalid or null `study_state`, plus a single valid case.

mozetl/shield/privacy_prefs.py
```diff
@@ -5,40 +5,85 @@
...
+def include_state_pings(ping):
+    return ping['payload/study_name'] == SHIELD_ADDON_ID
+
+
+def etl_job(sc, sqlContext, submission_date=None, save=True):
+    s3_path = 's3n://telemetry-parquet/harter/privacy_prefs_shield/v1'
+    if submission_date is None:
+        submission_date = (date.today() - timedelta(1)).strftime("%Y%m%d")
+
+    pings = Dataset.from_source(
```
acmiyaguchi: Nit: I think that using a new scope in parenthesis easier to read for long method chains.

```
(
    Dataset
    .from_source("telemetry")
    .where(
        docType="shield-study",
        submissionDate=submission_date,
        appName="Firefox",
    )
    .records(sc)
)
```

groovecoder:
Updated to remove redundant tests.

PR #102, opened 2017-08-17
Bug 1374831 - Sync bookmark validation

acmiyaguchi:
[Bug 1374831 - Schedule Sync bookmark validation job to run on Airflow](https://bugzilla.mozilla.org/show_bug.cgi?id=1374831)

This reformats [this gist](https://gist.github.com/kitcambridge/364f56182f3e96fb3131bf38ff648609) to run in this repository. [This runner notebook](https://gist.github.com/acmiyaguchi/5c0c1862a11849f7d54edd74386123c5) can be modified to run this notebook on ATMO. 

@kitcambridge r?

In particular, `bmk_total_per_day` is the most heavily affected by this change since it depends on unique clients per period. Counting uniques over several smaller datasets is more difficult.

This actually runs pretty quickly for a single day, it might be worthwhile to just run the query over a longer window and overwrite it weekly.


mozetl/sync/bookmark_validation.py
```diff
@@ -0,0 +1,133 @@
...
+    spark.conf.set("parquet.block.size", parquet_size)
+    spark.conf.set("dfs.blocksize", parquet_size)
+
+    version = 1
+    input_path = "s3://{}/{}".format(input_bucket, input_prefix)
+
+    # use the airflow date convention
+    ds_format = "YYYYMMDD"
+    start = arrow.get(start_date, ds_format)
+    end = arrow.get(end_date if end_date else start_date, ds_format)
```
kitcambridge: Just so I understand, will `start` and `end` always be the same day when running on Airflow?

acmiyaguchi: Yes, but if you set an end date, it'll run for multiple days.

acmiyaguchi:
Does this new definition of `bmk_total_per_day` work for you? I don't think you'll be able to aggregate it because of overlapping clients over multiple days (eg `SUM(total_validated_users) from bmk_total_per_day GROUP BY sync_day`), but you might be able to use it as some sort of rate (average number of validations per user per day)/.

kitcambridge:
I *think* that's OK. The three dashboards we'd like to build are the percent of validated users suffering from each problem, percent of checked bookmarks suffering from each problem, and error introduction for new users.

We're using the counts in `bmk_total_per_day` as denominators, so ISTM it's fine if we don't have all pings from a client for a particular day...we'll just count those problems and validations as part of the next day. I think that should still be enough to spot trends, match them up to bugs that we fixed, and identify improvements and regressions. Does that seem sensible to you?

You also mentioned we could overwrite the dataset weekly using a longer window...would that give us more accurate counts for older validations?

acmiyaguchi:
It's not necessarily more accurate, it just gives a different view of the data. It really depends on your use case. We can start from here and iterate as necessary -- the modifications should be straightforward.

acmiyaguchi:
Tagging @mreid-moz for technical review -- how do you feel about using spark tables as an intermediate data store? Also, do you think unit testing is appropriate here?

mozetl/sync/bookmark_validation.py
```diff
@@ -0,0 +1,133 @@
...
+@click.option('--prefix', default='kit/sync')
+@click.option('--input_bucket', default='telemetry-parquet')
+@click.option('--input_prefix', default='sync_summary/v1')
+def main(start_date, end_date, bucket, prefix, input_bucket, input_prefix):
+    spark = (SparkSession
+             .builder
+             .appName("sync_bookmark")
+             .getOrCreate())
+
+    # configure the spark session to partition the data reasonably
```
mreid-moz: I don't think this block size configuration has any effect (I think it's a holdover from old versions). Also, the `load` function collapses everything into 1 partition, so only one file will be written per table/`start_date` regardless of these block size directives. It should be safe to remove these lines.

mozetl/sync/bookmark_validation.py
```diff
@@ -0,0 +1,133 @@
...
+        path = (
+            "s3://{}/{}/{}/v{}/start_date={}"
+            .format(bucket, prefix, table_name, version, start_date)
+        )
+
+        logger.info("Saving table {} on start_date {} to {}"
+                    .format(table_name, start_date, path))
+
+        df = spark.sql("SELECT * FROM {}".format(table_name))
+        df.repartition(1).write.parquet(path, mode="overwrite")
```
mreid-moz: Do you know approximately how large the resulting parquet files will be with a single partition per `start_date`?

mozetl/sync/bookmark_validation.py
```diff
@@ -0,0 +1,133 @@
...
+        .createOrReplaceTempView("sync_summary")
+    )
+
+
+def transform(spark):
+    """Create the bookmark problem and summary tables."""
+
+    query = """
+    SELECT s.app_build_id, s.app_version, s.app_display_version, s.app_name,
+       s.app_channel, s.uid, s.deviceID as device_id,
```
mreid-moz: nit: s/as/AS/

mozetl/sync/bookmark_validation.py
acmiyaguchi: For `bmk_total_per_day` it's about 1kb. For `sync_validation_problems` it's 2.5kb.

acmiyaguchi:
@mreid-moz  RFAL

PR #108, opened 2017-09-04
Add the TAAR job to get locale-based recommendations

Dexterp37:
This PR adds a new Python job that builds a list of top addons per locale to use in the TAAR project.

Please do not review this just yet, I'm mainly opening the PR as I want to get tests running here (having problems locally).

### DIFF
--- a/mozetl/taar/taar_locale.py
+++ b/mozetl/taar/taar_locale.py
@@ -1,0 +1,182 @@
+"""
+Bug 1396549 - TAAR Top addons per locale dictionary
+This notebook is adapted from a gist that computes the top N addons per
+locale after filtering for good candidates (e.g. no unsigned, no disabled,
+...) [1].
+
+[1] https://gist.github.com/mlopatka/46dddac9d063589275f06b0443fcc69d
+"""
+
+import boto3
+import click
+import json
+import logging
+
+from pyspark.sql import SparkSession
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+LOCALE_FILE_NAME = "top10_dict"
+
+
+def store_new_state(source_file_name, s3_dest_file_name, s3_prefix, bucket):
+    """Store the new json file containing current top addons per locale to S3.
+
+    :param source_file_name: The name of the local source file.
+    :param s3_dest_file_name: The name of the destination file on S3.
+    :param s3_prefix: The S3 prefix in the bucket.
+    :param bucket: The S3 bucket.
+    """
+    client = boto3.client('s3', 'us-west-2')
+    transfer = boto3.s3.transfer.S3Transfer(client)
+
+    # Update the state in the analysis bucket.
+    key_path = s3_prefix + s3_dest_file_name
+    transfer.upload_file(source_file_name, bucket, key_path)
+
+
+def get_addons(spark):
+    """ Longitudinal sample is selected and freshest ping chosen per client.
+    Only Firefox release clients are considered.
+    Columns are exploded (over addon keys)  to include locale of each addon
+    installation instance system addons, disabled addons, unsigned addons
+    are filtered out.
+    Sorting by addon-installations and grouped by locale.
+    """
+    return spark.sql("""
+        WITH sample AS (
+        SELECT client_id,
+        settings[0].locale AS locality,
+        EXPLODE(active_addons[0])
+        FROM longitudinal
+        WHERE normalized_channel='release'
+          AND build IS NOT NULL
+          AND build[0].application_name='Firefox'
+        ),
+
+        filtered_sample AS (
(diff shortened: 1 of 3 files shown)


Dexterp37:
Hey @acmiyaguchi , I think this is ready for review now :) Would you mind taking a look at it?

mozetl/taar/taar_locale.py
```diff
@@ -0,0 +1,177 @@
...
+        from filtered_sample
+        GROUP BY locality, addon_key
+        )
+
+        SELECT * FROM country_addon_pairs
+        ORDER BY locality, pair_cnts DESC
+    """)
+
+
+def compute_threshold(addon_df):
```
acmiyaguchi: You could have calculated this using [`spark.sql.DataFrame.approxQuantile`](http://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.DataFrame.approxQuantile), which takes a column, a list of quantile probabilities in [0, 1], and a precision term, and returns a list of approximated quantiles.

mozetl/taar/taar_locale.py
```diff
@@ -0,0 +1,177 @@
...
+def store(dictionary, prefix, bucket):
+    FULL_FILENAME = "{}.json".format(LOCALE_FILE_NAME)
+
+    with open(FULL_FILENAME, "w+") as dict_file:
+        # If we attempt to load invalid JSON from the assembled file,
+        # the next function throws
+        dict_file.write(json.dumps(dictionary, indent=2))
+
+    archived_file_copy =\
+        "{}{}.json".format(LOCALE_FILE_NAME, datetime.today().strftime("%Y%d%m"))
```
acmiyaguchi: The date should be taken as an argument from the command line, to ensure idempotency. This allows for things like airflow backfill.

If you are not running on airflow, you could run this script manually (or scheduled via ATMO) through a script in this form.
```
# In[ ]:

# Stop the injected spark session from interfering with mozetl
spark.stop()

# In[ ]:

%%bash
cd /tmp
wget https://raw.githubusercontent.com/mozilla/python_mozetl/master/bin/mozetl-submit.sh
bash mozetl-submit.sh \
    -p https://github.com/mozilla/python_mozetl.git \
    taar_locale --start_date $(date -d '-1 day' '+%Y%m%d') \
        --foo value_1 \
        --bar value_2 
```

mozetl/taar/taar_locale.py
```diff
@@ -0,0 +1,177 @@
...
+    :param addon_df: the locale-specific addon dataframe;
+    :param threshold: the minimum number of addon-installs per locale;
+    :param num_addons: requested number of recommendations.
+    :return: a dictionary {<locale>: ['GUID1', 'GUID2', ...]}
+    """
+    top10_per = {}
+
+    # Decide that we can not make reasonable recommendations without
+    # a minimum number of addon installations.
+    grouped_addons = addon_df\
```
acmiyaguchi: Style nit: prefer implicit line continuations in parenthesis over backslashes, as per [pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). Here, it would look like this:

```
grouped_addons = (
    addon_df
    .groupBy('locality')
    .agg({'pair_cnts': 'sum'})
    .collect()
)

mozetl/taar/taar_locale.py
Dexterp37: Mh, good point. We just want the latest data for this job and don't care about older data (we just save a copy for future reference, but we're not planning to use it).

With that said, is ATMO scheduling preferred over Airflow scheduling in this case?

acmiyaguchi: No, you should probably use airflow (email notifications, retries). Mozetl in airflow essentially runs the second cell, but has a wrapper that passes context via environment variables instead of command line arguments.

If you're going to embed the date into the filename, I would still suggest taking the date from the environment. Ideally an ETL job should be able to run multiple times with the same date and always end up in the same state.

mozetl/taar/taar_locale.py
Dexterp37: Done!

mozetl/taar/taar_locale.py
Dexterp37: Good point! I found that `np.percentile` behaves slightly differently because it tries to interpolate the data points (instead of returning the closest one) if the requested percentile doesn't exactly match a data point ([docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html)). I had to tweak the test a little bit to account for that.

mozetl/taar/taar_locale.py
Dexterp37: Done :)

Dexterp37:
This is RFAL!

acmiyaguchi:
Looks good to me!

Issue #79, opened 2017-07-28
Add `pyspark` dependency (for spark 2.2.0+)


acmiyaguchi:
This requires the spark package on EMR to spark 2.2.0+

acmiyaguchi:
ATMO will be setting EMR 5.8.0 as default soon, which introduces spark 2.2.0. This means that we can avoid packaging a docker container for running tests.

acmiyaguchi:
Fixed with PR #109

PR #110, opened 2017-09-06
Fix issue #79 - Use pyspark via pip instead of docker (spark 2.2.0)

acmiyaguchi:
`pip` significantly reduces the complexity  of testing this repository. This only requires the system dependencies  (snappy and JRE) in order to run. This has the added benefit of fixing caching and code coverage issues,


harterrt:
Everything looks like it's working, and I do hate slow tests. A couple of questions:

It sounds like docker is no longer needed because pyspark will be installed by default on ATMO. Without this benefit, the extra test time isn't justified. Does that capture the rationale for this change? If so, will we ever be out of sync with ATMO again? Will we have to re-implement the docker tests when we upgrade to the next version of spark?

What code coverage issues will this fix?

acmiyaguchi:
ATMO is upgrading to EMR 5.8.0 soon meaning that the default spark version will be 2.2.0. This is also the version where pyspark is distributed via pip. Every subsequent major release will have a related pypy package. This makes the docker container obsolete since a Python package can be installed in a virtual environment that is independent of the system (at least in a unix like environment).

The Docker container and script were made to automate the spark installation and run tests. It was only a requirement if you wanted to run the unit tests on a local machine. The tests would run without Docker on ATMO because they already have spark and pyspark. 

The rationale is to get rid of all the unnecessary moving parts and to improve the speed of CI builds. That being said, the pyspark build should be pinned to the current ATMO version. We probably won't need to touch Docker again (unless you run on Windows), but the installation does require install libsnappy and the JRE.

Finally, we should be seeing more of the friendly `Codecov` bot in our PR requests.

PR #112, opened 2017-09-08
added specific blacklist to taar/taar_locale.py

mlopatka:
Added an early filter to remove addons (by key) that are notsuccessfully filtered using the "is_system" condition.


mozetl/taar/taar_locale.py
```diff
@@ -65,6 +65,11 @@ def get_addons(spark):
           AND value['app_disabled'] = FALSE -- exclude compatibility disabled addons
           AND value['is_system'] = FALSE -- exclude system addons
           AND locality <> 'null'
+          AND key <> 'loop@mozilla.org'
```
acmiyaguchi: You could use `key NOT IN ('loop@mozilla.org', 'firefox@getpocket.com', ...)`, to make it clear what the blacklist looks like.

acmiyaguchi:
It looks like this job is failing because approxQuantile is not as robust as it should be. This looks like it's [resolved in Spark 2.2.0](https://issues.apache.org/jira/browse/SPARK-21550). I've updated the test harness to use Spark 2.2.0, but it still needs to make its way into the tree (which *should* fix the problem here, unless the tests are broken in another way). This also means the airflow job should set 5.8.0 as the default version.

mozetl/taar/taar_locale.py
```diff
@@ -65,6 +65,8 @@ def get_addons(spark):
           AND value['app_disabled'] = FALSE -- exclude compatibility disabled addons
           AND value['is_system'] = FALSE -- exclude system addons
           AND locality <> 'null'
+          AND key NOT IN ('loop@mozilla.org', 'firefox@getpocket.com', 'e10srollout@mozilla.org',
```
Dexterp37: I think the right way to proceed here would be to check the AMO database: system, private addons and unlisted addons (e.g. the ones similar to the ubuntu system integration) will be discarded. We won't have to maintain yet-a-new blacklist and could re-use the logic in the other upcoming jobs. See [here](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/ml/AddonRecommender.scala#L147) for how we do this in the collaborative filtering job.

mlopatka: Yup, 100% agree, this is a much more elegant (and future-proof) solution.

I have added a commit. Please dont review just yet, test coverage comming soon :)

mozetl/taar/taar_locale.py
```diff
@@ -37,6 +40,12 @@ def store_new_state(source_file_name, s3_dest_file_name, s3_prefix, bucket):
     transfer.upload_file(source_file_name, bucket, key_path)
 
 
+def load_amo_external_whitelist():
+    amo_dump = utils.get_s3_json_content(AMO_DUMP_BUCKET, AMO_DUMP_KEY)
+    white_list_inner = {key: value['guid'] for key, value in amo_dump.items() if value['current_version']['files'][0]['is_webextension']}
```
Dexterp37: Fly-by: you should really check if any of the files supports webextension, not just the first one. For example:

```
    addon_files = value.get('current_version', {}).get('files', {})
    addon_data['is_webextension'] =\
        any([f.get("is_webextension", False) for f in addon_files])
```

mlopatka:
@acmiyaguchi Hey Anthony, can you take a quick peek at the PR?

tests/test_taar_locale.py
```diff
@@ -1,10 +1,11 @@
 """Test suite for TAAR Locale Job."""
 
 import boto3
+import json
 import functools
 import pytest
 from moto import mock_s3
-from mozetl.taar import taar_locale
+from mozetl.taar import taar_locale, AMO_DUMP_BUCKET, AMO_DUMP_KEY
```
acmiyaguchi: `AMO_DUMP_BUCKET` and `AMO_DUMP_KEY` are actually in the scope of `taar_locale`. You can access them like the functions e.g. `taar_locale.AMO_DUMP_BUCKET`

tests/test_taar_locale.py
```diff
@@ -130,3 +181,20 @@ def test_write_output():
...
+    content = [fake_amo_sample_web_extension, fake_amo_sample_legacy]
+
+    conn = boto3.resource('s3', region_name='us-west-2')
+    conn.create_bucket(Bucket=AMO_DUMP_BUCKET)
+
+    # Store the data in the mocked bucket.
+    conn.Object(AMO_DUMP_BUCKET, key=AMO_DUMP_KEY).put(Body=json.dumps(content))
+
+    # Check that the web_extension item is still present and the legacy addon is absent.
+    white_listed_addons = taar_locale.load_amo_external_whitelist()
```
acmiyaguchi: For some reason, it looks like the library calls to boto3 are not being mocked out by moto. When running this test locally, I am getting production data.

acmiyaguchi: This is because of the caching mechanism that is implemented in `get_s3_json_content`. Either the caching mechanism is removed (it only gets run in one place), or the test harness should clear the cache directory after every test.

mozetl/taar/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    The file is downloaded and then cached for future use.
+    """
+    local_path = get_s3_cache_filename(s3_bucket, s3_key)
+
+    if not os.path.exists(local_path):
+        # Use NamedTemporaryFile, so that the file gets removed.
+        with NamedTemporaryFile() as temp_file:
+            try:
+                s3 = boto3.client('s3')
+                s3.download_fileobj(s3_bucket, s3_key, temp_file)
```
acmiyaguchi: You could use `download_file` and avoid the use of the `NamedTemporaryFile` here, along with all of the other file level operations.

tests/test_taar_locale.py
```diff
@@ -130,3 +181,20 @@ def test_write_output():
     date_filename =\
         "{}{}20171106.json".format(prefix, taar_locale.LOCALE_FILE_NAME)
     assert date_filename in keys
+
+
+@mock_s3
+def test_load_amo_external_whitelist():
+
+    content = [fake_amo_sample_web_extension, fake_amo_sample_legacy]
```
acmiyaguchi: This content isn't in the right form, it looks like the database is one large json file that maps GUID to addon object.

mlopatka: The full AMO dump is transformed into the lost of guids inside load_amo_external_whitelist().
So I'm stubbing a full amo entry, but the output should be a list of guids.

tests/test_taar_locale.py
mlopatka: Ah... nice catch! I've been going crazy testing this. Thank you!

mozetl/taar/taar_locale.py
```diff
@@ -37,6 +41,24 @@ def store_new_state(source_file_name, s3_dest_file_name, s3_prefix, bucket):
     transfer.upload_file(source_file_name, bucket, key_path)
 
 
+def load_amo_external_whitelist():
+    white_list_inner = []
+    # Load the most current AMO dump JSON resource.
+    try:
+        amo_dump = utils.get_s3_json_content_no_cache(AMO_DUMP_BUCKET, AMO_DUMP_KEY)
```
Dexterp37: Instead of importing all the stuff in the `utils.py` file, which is not really needed here, why don't you do:

```
s3 = boto3.client('s3')
s3_contents = s3.get_object(Bucket=AMO_DUMP_BUCKET, Key=AMO_DUMP_KEY)
amo_dump = json.loads(s3_contents['Body'].read())
```

Within the try?

mozetl/taar/taar_locale.py
```diff
@@ -37,6 +40,26 @@ def store_new_state(source_file_name, s3_dest_file_name, s3_prefix, bucket):
     transfer.upload_file(source_file_name, bucket, key_path)
 
 
+def load_amo_external_whitelist():
+    white_list_inner = []
+    # Load the most current AMO dump JSON resource.
+    try:
+        s3 = boto3.client('s3')
+        s3_contents = s3.get_object(Bucket=AMO_DUMP_BUCKET, Key=AMO_DUMP_KEY)
+        amo_dump = json.loads(s3_contents['Body'].read())
```
Dexterp37: Let's define `amo_dump = {}` after line 44, this should make the test pass.

mlopatka: sorry about this.

Dexterp37:
Let's close this in favor of #116

PR #113, opened 2017-09-12
Bug 1381896 - add daywise rollups and search counts per access point

SamPenrose:
Per bug 1381896, convert main to run daywise rather than monthwise. Also add columns that tally searches for each chrome access point.

### DIFF
--- a/mozetl/clientsdaily/rollup.py
+++ b/mozetl/clientsdaily/rollup.py
@@ -3,20 +3,28 @@ import os
 from pyspark.sql import SparkSession
 from pyspark.sql import functions as F
 import click
-from moztelemetry.standards import filter_date_range
-from mozetl.utils import (
-    format_spark_path,
-    generate_filter_parameters
-)
+from mozetl.utils import format_spark_path
 from fields import MAIN_SUMMARY_FIELD_AGGREGATORS
 
-ACTIVITY_SUBMISSION_LAG = DT.timedelta(10)
+LAG = 10
+ACTIVITY_SUBMISSION_LAG = DT.timedelta(LAG)
+ACTIVITY_DATE_COLUMN = F.expr(
+    "substr(subsession_start_date, 1, 10)"
+).alias("activity_date")
 MAIN_SUMMARY_VERSION = 4
 MAIN_SUMMARY_PATH = "s3://telemetry-parquet/main_summary/v{}".format(
     MAIN_SUMMARY_VERSION)
-WRITE_VERSION = '3'
+WRITE_VERSION = '5'
 STORAGE_BUCKET = 'net-mozaws-prod-us-west-2-pipeline-analysis'
 STORAGE_PREFIX = '/spenrose/clients-daily/v{}/'.format(WRITE_VERSION)
+SEARCH_ACCESS_POINTS = [
+    'abouthome', 'contextmenu', 'newtab', 'searchbar', 'system', 'urlbar'
+]
+SEARCH_ACCESS_COLUMN_TEMPLATE = 'search_count_{}'
+SEARCH_ACCESS_COLUMNS = [
+    SEARCH_ACCESS_COLUMN_TEMPLATE.format(sap)
+    for sap in SEARCH_ACCESS_POINTS
+]
 
 
 def load_main_summary(spark):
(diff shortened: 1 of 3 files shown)


mozetl/clientsdaily/rollup.py
```diff
@@ -10,13 +10,25 @@
 )
 from fields import MAIN_SUMMARY_FIELD_AGGREGATORS
 
-ACTIVITY_SUBMISSION_LAG = DT.timedelta(10)
+LAG = 10
+ACTIVITY_SUBMISSION_LAG = DT.timedelta(LAG)
+ACTIVITY_DATE_COLUMN = F.expr(
+    "substr(subsession_start_date, 1, {})".format(LAG)
```
mreid-moz: This shouldn't be using `LAG` - it's just chopping off the date prefix of the date string, right?

SamPenrose: Doh! How silly of me.

mreid-moz: Easy mistake to make when extracting a constant (which is a good thing to do).

mozetl/clientsdaily/rollup.py
```diff
@@ -30,23 +42,74 @@ def load_main_summary(spark):
 
 def extract_search_counts(frame):
```
acmiyaguchi: The [pivot operation](https://databricks.com/blog/2016/02/09/reshaping-data-with-pivot-in-apache-spark.html) might reduce some of the complexity here.  Something along the lines of `groupBy("did").pivot("search_struct.source", SEARCH_ACCESS_POINTS).agg(...)`

mozetl/clientsdaily/rollup.py
```diff
@@ -79,15 +142,38 @@ def extract_month(first_day, frame):
...
+    :day DT.date(Y, m, d)
+    :frame DataFrame homologous with main_summary
+    """
+    frame = frame.select("*", ACTIVITY_DATE_COLUMN)
+    activity_iso = day.isoformat()
+    activity_submission_str = day.strftime('%Y%m%d')
+    submission_end = day + ACTIVITY_SUBMISSION_LAG
+    submission_end_str = submission_end.strftime('%Y%m%d')
+    result = frame.where(
+        "submission_date_s3 >= '{}'".format(activity_submission_str)) \
```
acmiyaguchi: Nit: The backslashes aren't necessary

mozetl/clientsdaily/rollup.py
```diff
@@ -97,10 +183,23 @@ def write_by_activity_day(
         prefix = prefix_template.format(isoday)
         output_path = format_spark_path(output_bucket, prefix)
         data_for_date = results.where(results.activity_date == isoday)
-        data_for_date.coalesce(partition_count).write.parquet(output_path)
+        to_write = data_for_date.coalesce(partition_count)
+        to_write.write.parquet(output_path)
+        to_write.unpersist()
```
acmiyaguchi: Do you need to unpersist if you never persist the dataframe?

mozetl/clientsdaily/rollup.py
```diff
@@ -114,6 +213,36 @@ def get_partition_count_for_writing(is_sampled):
     return 25
 
 
+def monthwise_main(date, input_bucket, input_prefix, output_bucket,
```
acmiyaguchi: Where is this getting called?

mozetl/clientsdaily/rollup.py
SamPenrose: Your argument is with flake8 sir.

mozetl/clientsdaily/rollup.py
SamPenrose: Doc for unpersist(): "Marks the DataFrame as non-persistent, and remove all blocks for it from memory and disk." It's possible the implicit __del__ as the objects are unbound will be as fast, but maybe not. Also these lines emphasize how tight memory is on our clusters during this process.

mozetl/clientsdaily/rollup.py
SamPenrose: It's not. I left the monthwise code in case someone wanted to use it. If y'all would rather I delete it that's fine by me.

acmiyaguchi: Do you think this would be the best way to perform backfill on large batches of `main_summary`? It could be useful when clients-daily gets ramped up from 1%.

SamPenrose: I think at 100% monthly almost certainly won't work due to exponential growth in the work done by groupBy(client_id), so come to think of it we should probably just delete it.

PR #115, opened 2017-09-16
Taar similarity job

Dexterp37:
This job is used to compute TAAR similarity-based addon donor list.

This job clusters users into different groups based on their active addons. A representative users sample is selected from each cluster ("donors") and is saved to a model file along with a feature vector that will be used, by the TAAR library module, to perform recommendations.

*Please do not review yet*

### DIFF
--- a/mozetl/taar/taar_similarity.py
+++ b/mozetl/taar/taar_similarity.py
@@ -1,0 +1,330 @@
+"""
+Bug 1386274 - TAAR similarity-based add-on donor list
+
+This job clusters users into different groups based on their
+active add-ons. A representative users sample is selected from
+each cluster ("donors") and is saved to a model file along
+with a feature vector that will be used, by the TAAR library
+module, to perform recommendations.
+"""
+
+import click
+import json
+import logging
+import numpy as np
+
+from pyspark.sql import SparkSession
+from pyspark.sql.functions import col
+from pyspark.ml.feature import HashingTF, IDF
+from pyspark.ml.clustering import BisectingKMeans
+from pyspark.ml import Pipeline
+from pyspark.mllib.stat import KernelDensity
+from scipy.spatial import distance
+from taar_utils import store_json_to_s3, load_amo_external_whitelist
+
+# Define the set of feature names to be used in the donor computations.
+CATEGORICAL_FEATURES = ["geo_city", "locale", "os"]
+CONTINUOUS_FEATURES = \
+    ["subsession_length", "bookmark_count", "tab_open_count", "total_uri", "unique_tlds"]
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_samples(spark):
+    """ Get a DataFrame with a valid set of sample to base the next
+    processing on.
+    """
+    return (
+        spark.sql("SELECT * FROM longitudinal")
+        .where("active_addons IS NOT null")
+        .where("size(active_addons[0]) > 2")
+        .where("size(active_addons[0]) < 100")
+        .where("normalized_channel = 'release'")
+        .where("build IS NOT NULL AND build[0].application_name = 'Firefox'")
+        .selectExpr(
+            "client_id as client_id",
+            "active_addons[0] as active_addons",
+            "geo_city[0] as geo_city",
+            "subsession_length[0] as subsession_length",
+            "settings[0].locale as locale",
+            "os as os",
+            "places_bookmarks_count[0].sum AS bookmark_count",
+            "scalar_parent_browser_engagement_tab_open_event_count[0].value AS tab_open_count",
+            "scalar_parent_browser_engagement_total_uri_count[0].value AS total_uri",
+            "scalar_parent_browser_engagement_unique_domains_count[0].value AS unique_tlds"
+        )
+    )
+
+
+def get_addons_per_client(users_df, addon_whitelist, minimum_addons_count):
+    """ Extracts a DataFrame that contains one row
+    for each client along with the list of active add-on GUIDs.
+    """
+    def is_valid_addon(guid, addon):
+        return not (
+            addon.is_system or
+            addon.app_disabled or
+            addon.type != "extension" or
+            addon.user_disabled or
+            addon.foreign_install or
+            guid not in addon_whitelist
+        )
+
+    # Create an add-ons dataset un-nesting the add-on map from each
+    # user to a list of add-on GUIDs. Also filter undesired add-ons.
+    return (
+        users_df.rdd
+        .map(lambda p: (p["client_id"],
+             [guid for guid, data in p["active_addons"].items() if is_valid_addon(guid, data)]))
+        .filter(lambda p: len(p[1]) > minimum_addons_count)
+        .toDF(["client_id", "addon_ids"])
+    )
+
+
+def compute_clusters(addons_df, num_clusters, random_seed):
+    """ Performs user clustering by using add-on ids as features.
+    """
+
+    # Build the stages of the pipeline. We need hashing to make the next
+    # steps work.
+    hashing_stage = HashingTF(inputCol="addon_ids", outputCol="hashed_features")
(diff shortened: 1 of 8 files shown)


Dexterp37:
@mlopatka, can you take an high level look to check that the similarity based module behaves as expected? Please also check that the test is testing the right thing.

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,163 @@
...
+
+def get_samples(spark):
+    """ Get a dataframe with a valid set of sample to base the next
+    processing on.
+    """
+    return spark.sql("SELECT * FROM longitudinal")\
+        .where("active_addons IS NOT null")\
+        .where("size(active_addons[0]) > 2")\
+        .where("size(active_addons[0]) < 100")\
+        .where("normalized_channel = 'release'")\
```
mlopatka: Don't we also want:
` AND build IS NOT NULL AND build[0].application_name='Firefox'`

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,163 @@
...
+
+def compute_clusters(addons_df, num_clusters):
+    """ Performs user clustering by using addon ids as features.
+    """
+
+    # Build the stages of the pipeline. We need hashing to make the next
+    # steps work.
+    hashing_stage = HashingTF(inputCol="addon_ids", outputCol="hashed_features")
+    idf_stage = IDF(inputCol="hashed_features", outputCol="features", minDocFreq=1)
+    bkmeans_stage = BisectingKMeans(k=num_clusters)
```
mlopatka: Not necessary now, but a future imporvement may be to have the termination condition relate to `minDivisibleClusterSize`. Since Bisecting K-means is a divisive clustering method, we may want to continue to sub-divide the data until the smallest single cluster contains a minimum number of elements.

tests/test_taar_similarity.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def test_compute_donors(spark, multi_clusters_df):
+    multi_clusters_df.createOrReplaceTempView("longitudinal")
+
+    # Perform the clustering on our test data. We expect
+    # 3 clusters out of this and 10 donors.
+    donors = taar_similarity.get_donors(spark, 3, 10)
+
+    # assert len(donors) == 10
```
mlopatka: There should be an assertion that the number of donors is within a particular margin of the requested number. It will not be equal due to the sampling from variable sized donor candidates (accross clusters) however the number of donors should always be within 2*(n) where n is the number of terminal clusters.

mozetl/taar/taar_similarity.py
Dexterp37: Good catch!

mozetl/taar/taar_similarity.py
Dexterp37: Yup, I've added a comment line there.

tests/test_taar_similarity.py
Dexterp37: Good point. However, to make tests less flaky, I've slightly changed the code to pass around a fixed "random seed" to the Spark ML functions to make tests reproducible. Since we generate the fake data, this allows us to precisely assert how many donors we see.

mlopatka: Ah yeah, good point. For generated data it is a deterministic process, given a fixed random seed. Maybe we should add logging in the live module that warns us if this condition is not met so we can assess variability in with the cluster sizes as new client data is pulled in.

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Sample the users in each cluster according to the proportions
+    # and pass along the random seed if needed for tests.
+    sampling_kwargs = {"seed": random_seed} if random_seed else {}
+    donor_df = clusters_df.sampleBy("prediction",
+                                    fractions=sampling_proportions,
+                                    **sampling_kwargs)
+    # Get the specific number of donors for each cluster and drop the
+    # predicted cluster number information.
+    current_sample_size = donor_df.count()
```
mlopatka: Can we add logging here to check if the actual number of returned donors is close to the requested number of donors?

Dexterp37: @mlopatka yes, I will add the logging line here. I have a question though: should we make the job fail if we don't get enough donors?

mlopatka: As we just discusssed, we need a hard-coded failure condition for the extreme case.
The threshold should be set to <100.

Likewise, if a number of donoros less than 100 is requested at any funciton call (as input param) the job shoudl fail as well.

mlopatka:
@acmiyaguchi Can you take a look at this PR for the final r+?

Dexterp37:
@mlopatka @acmiyaguchi this is not yet ready for review.

Dexterp37:
Please note that I could now write test coverage for the lr -curves function: it consistently timed out locally in the test unless I increased the driver memory to over 2 GB.

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,321 @@
...
+                  num_pdf_points, random_seed=None):
+    """ Compute the likelihood ratio curves for clustered clients.
+
+    Work-flow followed in this function is as follows:
+
+     * Access the DataFrame including cluster numbers and features.
+     * Load same similarity function that will be used in TAAR module.
+     * Iterate through each cluster and compute in-cluster similarity.
+     * Iterate through each cluster and compute out-cluster similarity.
+     * Compute KDE density per similarity score.
```
mlopatka: nit: KDE can be written 

> Kernel Density Estimate

then we can drop the redundant "density"

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,321 @@
...
+    x_continuous_features = [x[k] for k in CONTINUOUS_FEATURES]
+    y_categorical_features = [y[k] for k in CATEGORICAL_FEATURES]
+    y_continuous_features = [y[k] for k in CONTINUOUS_FEATURES]
+
+    # Here a larger distance indicates a poorer match between categorical variables.
+    j_d = (distance.hamming(x_categorical_features, y_categorical_features))
+    j_c = (distance.canberra(x_continuous_features, y_continuous_features))
+
+    # Take the product of similarities to attain a univariate similarity score.
+    # Add a minimal constant to prevent zero values.
```
mlopatka: nit: to prevent zero values `from categorical features`.

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,321 @@
...
+        )
+    )
+
+
+def get_addons_per_client(users_df, minimum_addons_count):
+    """ Extracts a DataFrame that contains one row
+    for each client along with the list of active add-on GUIDs.
+    """
+    def is_valid_addon(addon):
+        if addon.is_system or\
```
acmiyaguchi: Nit: You could wrap this in parens and return the negation.

mozetl/taar/taar_similarity.py
```diff
@@ -0,0 +1,321 @@
...
+        return True
+
+    # Create an add-ons dataset un-nesting the add-on map from each
+    # user to a list of add-on GUIDs. Also filter undesired add-ons.
+    return (
+        users_df.rdd
+        .map(lambda p: (p["client_id"],
+             [guid for guid, data in p["active_addons"].items() if is_valid_addon(data)]))
+        .filter(lambda p: len(p[1]) > minimum_addons_count)
+        .toDF()
```
acmiyaguchi: I think you can pass the column names into `.toDF`

Dexterp37:
@acmiyaguchi thanks for the review. I've addressed your points and also fixed the `similarity_function` to be more robust against missing values in the data. Moreover, since the locale PR landed, I moved the whitelist code out of that module and used it to filter the addons in this module as well. Can you take another look at the changed parts?

acmiyaguchi:
@Dexterp37 The changes look good, looks like all of your changes have all the interesting cases being exercised by relevant tests.

PR #116, opened 2017-09-19
Restrict the recommended add-ons to the ones in AMO for taar_locale

Dexterp37:
AMO will only report compatible, public add-ons. OS widgets and other unlisted/unapproved add-ons are not reported and will filter the recommendation list.

### DIFF
--- a/mozetl/taar/taar_locale.py
+++ b/mozetl/taar/taar_locale.py
@@ -12,13 +12,16 @@ import click
 import json
 import logging
 
+from botocore.exceptions import ClientError
 from pyspark.sql import SparkSession
+from pyspark.sql.functions import col
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
-
-LOCALE_FILE_NAME = "top10_dict"
+LOCALE_FILE_NAME = 'top10_dict'
+AMO_DUMP_BUCKET = 'telemetry-parquet'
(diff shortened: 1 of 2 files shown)


mozetl/taar/taar_locale.py
```diff
@@ -37,6 +40,29 @@ def store_new_state(source_file_name, s3_dest_file_name, s3_prefix, bucket):
...
+        # Load the most current AMO dump JSON resource.
+        s3 = boto3.client('s3')
+        s3_contents = s3.get_object(Bucket=AMO_DUMP_BUCKET, Key=AMO_DUMP_KEY)
+        amo_dump = json.loads(s3_contents['Body'].read())
+    except ClientError:
+        logger.exception("Failed to download from S3", extra={
+            "bucket": AMO_DUMP_BUCKET,
+            "key": AMO_DUMP_KEY})
+
+    # If the load fails, we will have an empty whitelist, this may be problematic.
```
acmiyaguchi: If the whitelist fails to load, should the job fail or continue?

Dexterp37: There's an assert on line 193: do you want me to do something different/move it?

acmiyaguchi: Can you raise an [appropriate exception](https://docs.python.org/2/library/exceptions.html) instead of the assertion? It can be here or in 193.

acmiyaguchi:
With the change to raising an exception, LGTM

Dexterp37:
Thanks Anthony!

PR #117, opened 2017-09-20
first take of experiments-daily with broken tests

SamPenrose:
Per bug 1401646, this contains the core experiments-daily rollup, which handles the entire (small) dataset on each run. The test is broken because of some local PySpark weirdness. This statement in clients_daily.rollup.extract_searches:

```
    exploded = two_columns.select(
        "did", F.explode("search_counts").alias("search_struct"))
```

returns an empty Dataframe for some reason instead of one holding about two dozen rows.


mozetl/experimentsdaily/fields.py
```diff
@@ -0,0 +1,173 @@
+import pyspark.sql.functions as F
+EXPERIMENT_FIELD_AGGREGATORS = [
```
acmiyaguchi: Can you reuse ` mozetl/clientsdaily/fields.py` data? It looks like the only differences are:

```
-  F.first('active_experiment_branch').alias('active_experiment_branch'),
-  F.first('active_experiment_id').alias('active_experiment_id'),
+  F.first('experiment_branch').alias('experiment_branch'),
```

mozetl/experimentsdaily/rollup.py
```diff
@@ -0,0 +1,73 @@
...
+STORAGE_PREFIX = '/spenrose/experiments-daily/v{}/'.format(WRITE_VERSION)
+
+
+def load_experiments_summary(spark, parquet_path):
+    return (
+        spark
+        .read
+        .option("mergeSchema", "true")
+        .parquet(parquet_path)
+        .where("experiment_id != '{}'".format(EXCLUDED_ID))
```
acmiyaguchi: Should you also exclude all rows/clients that are not enrolled in an experiment?

tests/test_experimentsdaily.py
```diff
@@ -0,0 +1,46 @@
...
+    frame = spark.read.json(rows_path, schema)
+    return frame
+
+
+def test_rollup(spark):
+    from mozetl.experimentsdaily import rollup
+    from mozetl.clientsdaily.rollup import extract_search_counts
+    frame = make_frame(spark)
+    searches_frame = extract_search_counts(frame)
+    return
```
acmiyaguchi: Please use [the @pytest.mark.skip(reason="...") decorator](https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions) to skip this test.

mozetl/experimentsdaily/rollup.py
SamPenrose: s3://telemetry-parquet/experiments/v1/ is all main_summary rows that are not in an experiment.

acmiyaguchi: It's the route for the ExperimentSummaryView ([telemetry.experiments sample query](https://sql.telemetry.mozilla.org/queries/39852/source)), which is all the main_summary rows that are in an experiment, grouped by experiment and submission date.

Does the output of this script keep clients that do not have an `experiment_id`?

tests/test_experimentsdaily.py
```diff
@@ -0,0 +1,47 @@
...
+
+@pytest.mark.skip(reason="extract_search_counts broken locally")
+def test_rollup(spark):
+    from mozetl.experimentsdaily import rollup
+    from mozetl.clientsdaily.rollup import extract_search_counts
+    frame = make_frame(spark)
+    searches_frame = extract_search_counts(frame)
+
+    # It's actually 0 due to something in local PyPspark
+    assert searches_frame.count() == 100
```
acmiyaguchi: Change this to the value it *should* be.

PR #120, opened 2017-09-24
Make the similarity job work around SPARK-20803

Dexterp37:
Without this fix, the job will fail on Spark 2.2.0 due to [a bug](https://issues.apache.org/jira/browse/SPARK-20803) in the Kernel Density Estimation not correctly handling Numpy types. Moreover, the job will go OOM on previous Spark versions due to other bugs, so we can't really use anything less than Spark 2.2.0.

This fix was tested on EMR 5.8.0 and the results are accessible in the test telemetry bucket under the TAAR directory

### DIFF
--- a/mozetl/taar/taar_similarity.py
+++ b/mozetl/taar/taar_similarity.py
@@ -199,7 +199,11 @@ def similarity_function(x, y):
 
     # Take the product of similarities to attain a univariate similarity score.
     # Add a minimal constant to prevent zero values from categorical features.
-    return abs((j_c + 0.001) * j_d)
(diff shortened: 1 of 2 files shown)


mlopatka:
This looks like it generates properly formed and mostly expected output.
The analysis of the LR curves is [here](https://gist.github.com/mlopatka/50afbf8eb0f262b750e829e3e004c209) 

Int he future we may want to increase the number of clusters before pairwise comparison is performed to avoid the high LR values at higher scores (>3.0). Int he mean time this is sufficient for launching the similarity recommender and should give expected outcome given the current parameter choices in the TAAR module. 

Amazing job Alessio!

PR #121, opened 2017-09-24
Add a utility function for reading main_summary

mreid-moz:
This supports merging schemas for only a subset of partitions.

Test data was generated using [this script](https://gist.github.com/mreid-moz/14f177f461486692af945176c1ede87d)


fbertsch:
The problem with this approach is we lose the `submission_date_s3` and `sample_id` columns. With `main_summary` we do have `submission_date`, which may make this acceptable; we could also add `sample_id_np` to the data (for `sample_id non-partitioned`) so that `sample_id` is still available for new data. Most use cases will not require `sample_id`, however, so it could just be an acceptable loss.

If we wanted to include them, we could read each partition as a `DataFrame`, add the partitioned columns with a `withColumn` statement, and `union` them together.

We should probably add a test case to show this as well; could be in `test_base_path`.

mreid-moz:
Closing this in favour of [this PR in python_moztelemetry](https://github.com/mozilla/python_moztelemetry/pull/177).

PR #123, opened 2017-09-25
Clean up experimentsdaily code to run via Airflow.

mreid-moz:
We need to avoid manipulating Spark SQL functions before setting up the Spark context, and I removed a few unused arguments.


mreid-moz:
We may want to change the API to offer a more explicit "spark context is ready, do your initialization" step, but we might also change the approach in this code based on [Bug 1384200](https://bugzilla.mozilla.org/show_bug.cgi?id=1384200).

PR #124, opened 2017-09-26
Bug 1389231 - Add new-profile ping to churn

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1389231)

This adds the `new-profile` ping as a data source for churn. A new
column `is_new_profile` differentiates the source for the clients.

This is also a version bump to `v3`.

Additionally, `spark.sql.session.timeZone` is set to UTC, which require
Spark 2.2.0 to work (see SPARK-18936 for details). The spark session
assumes the timezone of the local machine, which can cause failing
tests.

### DIFF
--- a/mozetl/churn/churn.py
+++ b/mozetl/churn/churn.py
@@ -145,12 +145,10 @@ def extract(main_summary, new_profile, start_ds, period, slack, is_sampled):
         main_summary
         .where(reduce(operator.__and__, predicates))
         .select(SOURCE_COLUMNS)
-        .withColumn("is_new_profile", F.lit(False))
         .union(
             clean_new_profile(new_profile)
             .where(reduce(operator.__and__, predicates))
             .select(SOURCE_COLUMNS)
-            .withColumn("is_new_profile", F.lit(True))
         )
     )
 
@@ -258,7 +256,6 @@ def clean_columns(prepared_clients, effective_version, start_ds):
     is_funnelcake = F.col('distribution_id').rlike("^mozilla[0-9]+.*$")
 
     attr_mapping = {
-        'is_new_profile': None,
         'distribution_id': None,
         'default_search_engine': None,
         'locale': None,
--- a/mozetl/churn/churn.py
+++ b/mozetl/churn/churn.py
@@ -88,11 +88,6 @@ def clean_new_profile(new_profile):
 
     iso8601_tz_format = "yyyy-MM-dd'T'HH:mm:ss.S'+00:00'"
 
-    subsession_length = (
-        F.col("metadata.timestamp") -
-        F.col("metadata.creation_timestamp")
-    ) / 10**9
-
     select_expr = {
         "app_version": "environment.build.version",
         "attribution": "environment.settings.attribution",
@@ -103,11 +98,7 @@ def clean_new_profile(new_profile):
         "locale": "environment.settings.locale",
         "normalized_channel": "metadata.normalized_channel",
         "profile_creation_date": "environment.profile.creation_date",
-        "subsession_length": (
-            F.when(subsession_length < 0, 0)
-            .otherwise(subsession_length)
-            .cast('long')
-        ),
+        "subsession_length": F.lit(None).cast('long'),
         # The subsession_start_date is the profile_creation_date
         "subsession_start_date": F.from_unixtime(
             F.col("metadata.creation_timestamp") / 10 ** 9, iso8601_tz_format),
--- a/mozetl/churn/churn.py
+++ b/mozetl/churn/churn.py
@@ -61,7 +61,7 @@ SOURCE_COLUMNS = [
     "sync_count_mobile",
     "timestamp",
     "scalar_parent_browser_engagement_total_uri_count",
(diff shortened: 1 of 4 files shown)


mozetl/churn/churn.py
```diff
@@ -79,12 +79,50 @@
...
+        "subsession_start_date": F.from_unixtime(
+            F.col("metadata.creation_timestamp") / 10 ** 9, iso8601_tz_format),
+        # no access to `WEAVE_CONFIGURED`, `WEAVE_DEVICE_COUNT_*` histograms in the new_profile
+        "sync_configured": F.lit(None).cast("boolean"),
+        "sync_count_desktop": F.lit(None).cast("int"),
+        "sync_count_mobile": F.lit(None).cast("int"),
+        "timestamp": F.col("metadata.timestamp"),
+        "scalar_parent_browser_engagement_total_uri_count": F.lit(None).cast("int"),
+        "scalar_parent_browser_engagement_unique_domains_count": F.lit(None).cast("int"),
+        "sample_id": F.expr("crc32(encode(client_id, 'UTF-8')) % 100"),
```
mreid-moz: Please add a test case that ensures the `sample_id` computation matches with values from main_summary for a few client_ids. If you want you can use mine:
`client_id = 'c4582ba1-79fc-1f47-ae2a-671118dccd8b', sample_id = 4`

mozetl/churn/churn.py
```diff
@@ -107,6 +145,13 @@ def extract_subset(main_summary, start_ds, period, slack, is_sampled):
         main_summary
         .where(reduce(operator.__and__, predicates))
         .select(SOURCE_COLUMNS)
+        .withColumn("is_new_profile", F.lit(False))
+        .union(
+            clean_new_profile(new_profile)
+            .where(reduce(operator.__and__, predicates))
+            .select(SOURCE_COLUMNS)
+            .withColumn("is_new_profile", F.lit(False))
```
mreid-moz: s/False/True/

tests/churn/test_churn.py
```diff
@@ -181,27 +249,54 @@ def multi_profile_df(generate_data):
...
+    df = churn.extract(
+        generate_main_summary_data(None),
+        spark.createDataFrame([], new_profile_schema),
+        week_start_ds, 1, 0, False
+    )
     assert df.count() == 1
 
 
-def test_ignored_submissions_outside_of_period(generate_data):
+def test_extract_new_profile(spark, generate_new_profile_data):
```
mreid-moz: Please add a test for `extract` that uses both `main_summary` and `new_profile`. Ideally covering both the case where `new_profile` contains the same `client_id`s as well as different `client_id`s

acmiyaguchi: Good call, I realized that the `is_new_profile` column can't be used to distinguish the source of the clients because a client can be in both categories. This shouldn't be a problem with HLL, but it can't be used here without also including a "all" value.

acmiyaguchi:
@mreid-moz RFAL, but PR #130 takes precedence.

mreid-moz:
Since the `is_new_profile` field doesn't reliably map to the data we had before (since it depends what order we received the main vs. new-profile submissions), should we just drop that field from the output?

tests/churn/test_churn.py
```diff
@@ -123,27 +189,29 @@ def generate_dates(subsession_date, submission_offset=0, creation_offset=0):
...
         subsession_start.replace(days=3), creation_offset=3)
 
     # create a duplicate ping for this user, earlier than the previous
     old_ping = generate_dates(subsession_start)
 
     snippets = [recent_ping, old_ping]
-    return generate_data(snippets)
+    return (
+        generate_main_summary_data(snippets)
+        .withColumn("is_new_profile", F.lit(False))
```
mreid-moz: Don't need this field anymore.

tests/churn/test_churn.py
```diff
@@ -181,27 +249,89 @@ def multi_profile_df(generate_data):
     })
 
     snippets = [user_0, user_1, user_2]
-    return generate_data(snippets)
+    return (
+        generate_main_summary_data(snippets)
+        .withColumn("is_new_profile", F.lit(False))
```
mreid-moz: Don't need this field anymore

tests/churn/test_churn.py
```diff
@@ -256,13 +386,17 @@ def test_cohort_by_channel_aggregates(multi_profile_df, effective_version):
...
 
 @pytest.fixture()
-def test_transform(generate_data, effective_version):
+def test_transform(generate_main_summary_data, effective_version):
     def _test_transform(snippets, week_start=week_start_ds):
         return churn.transform(
-            generate_data(snippets),
+            (
+                generate_main_summary_data(snippets)
+                .withColumn("is_new_profile", F.lit(False))
```
mreid-moz: Don't need this field anymore.

mreid-moz:
Oh, also please update the main PR comment to reflect that `is_new_profile` got dropped.

PR #126, opened 2017-09-26
Make experimentsdaily tests run

mreid-moz:
Note that the first commit simply reformats the previous schema so that the subsequent diffs are easier to identify.


mreid-moz:
baby steps :)

PR #128, opened 2017-09-27
Bug 1401655 - Prepare jobs to run via Airflow.

mreid-moz:
In particular, change the `clientsdaily` job to accept an "end" date
instead of a "start" date to simplify the airflow side. Also accept
the LAG specification as an argument.


acmiyaguchi:
The failed tests are from a recent version of `moto` missing static resources. This should be good to merge in.

PR #129, opened 2017-09-27
Add sanity checks to hardware-report output

fbertsch:
These checks for % changes in values. If it exceeds a threshold,
(set to 30% change or more), then it emails us about the change.

Confirmed that emailing works as intended on an ATMO cluster.


mozetl/hardware_report/check_output.py
```diff
@@ -0,0 +1,61 @@
...
+    return {int(re.sub('-', '', e['date'])): {k: e[k] for k in e if k != 'date'} for e in data}
+
+
+def _check_most_recent_change(values, min_change=.05, min_value=0.01, missing_val=.01):
+    recent_week = max(values.keys())
+    second_recent_week = max(values.viewkeys() - {recent_week})
+
+    base, compare = values[second_recent_week], values[recent_week]
+    changes = [
+        (k, (compare.get(k, missing_val) - base.get(k, missing_val)) / base.get(k, missing_val))
```
harterrt: Is zero division possible here (in practice)? Maybe we should break this comparison out into a separate function or filter to minimum base size before calculating the change.

mozetl/hardware_report/check_output.py
```diff
@@ -0,0 +1,61 @@
+import ujson
+import re
+
+from mozetl.utils import send_ses
+
+
+def check_output():
+    data = _get_data()
+    changes = _check_most_recent_change(data, min_change=3)
```
harterrt: You mention 30% in a review comment. Should this be 0.3 instead of 3?

mozetl/hardware_report/check_output.py
harterrt: Totally optional: using `(compare/base) - 1` may be more succinct.

fbertsch:
@harterrt Thanks for the review! Added an assertion to ensure we never end up dividing by zero. Thanks for catching the `.3` error :)

PR #133, opened 2017-09-29
Bug 1381840 - Add 1 day Firefox retention intermediate dataset

acmiyaguchi:
In its current state, this should be considered `retention_dev_v3`. The
link to the STMO example dashboard for related datasets can be found in
[1].

This contains the cleaning routines for `main_summary` and `new_profile`
for use in a 1 day retention day dataset. There can be several instances
of the same client across a day (on average ~3 pings per client). This
is later processed by aggregating using HyperLogLog UDF's in
Telemetry-Batch-View.

This contains necessary columns for calculating retention by
attribution over profile counts and for calculating engagement over
subpopulations.

The extract routine is borrowed from churn, while transform uses the
same subroutines for cleaning and transforming the dataset.

[1] https://sql.telemetry.mozilla.org/dashboard/firefox-telemetry-retention-dataset-example-usage

@mreid f?


acmiyaguchi:
This is missing a few crucial bits before it can be merged in, but most of the logic is here.

* A test suite
* Documenting data cleaning goals
* Restricting clients to Firefox 55+

The first should be quick work since most of the test harness from the `churn_v2` refactor can be used in this dataset. The second item can go either way since most of the older clients would fall into the `NOT is_active` bucket.

acmiyaguchi:
The newest commit here reorganizes the project a bit to include an `engagement` module. This includes `churn` and `retention` so far. The two projects share a fair bit of code, both in the module and the tests. Additionally, relevant tests are added to the daily retention. The new folder structure is as follows:

```
mozetl/engagement/
├── churn
│   ├── __init__.py
│   ├── main.py
│   ├── release.py
│   ├── schema.py
│   └── utils.py
├── __init__.py
└── retention
    ├── __init__.py
    ├── main.py
    └── schema.py

tests/engagement/
├── conftest.py
├── data.py
├── __init__.py
├── test_churn.py
├── test_churn_release.py
├── test_churn_utils.py
└── test_daily_retention.py
```

The `__init__.py` file imports the relevant `main` function into the module scope, so the following bit of code can be run from the cli.

```
import mozetl.engagement.retention
retention.main()
```

I'd like to rename the file within the module to something other than `main` to make `from mozetl.engagement.retention import main` less ambiguous. Any suggestions?

mozetl/engagement/churn/__init__.py
```diff
@@ -1 +1 @@
-from main import *  # NOQA
+from job import *  # NOQA
```
jezdez: Ugh, this is a red flag, start imports are considered an anti-pattern since they hide the underlying implementation and confuse debuggers. If it's just to save typing "job" in import statements, I think it's fair to stop providing this shim.

mozetl/engagement/retention/__init__.py
```diff
@@ -1 +1 @@
-from main import *  # NOQA
+from job import *  # NOQA
```
jezdez: Another red flag.

tests/engagement/test_churn.py
```diff
@@ -1,207 +1,23 @@
...
 
 import pytest
 from pyspark.sql import functions as F
-from pyspark.sql.types import (
-    StructField, StructType, StringType,
-    LongType, IntegerType, BooleanType
-)
 
-from mozetl.churn import churn, schema
+from data import SUBSESSION_START, WEEK_START_DS, generate_dates, \
```
jezdez: Importing from `data` is an implicit relative import that are discouraged (and won't work on Python 3.x), use instead `from .data import ....`.

mozetl/cli.py
```diff
@@ -15,6 +16,7 @@ def entry_point():
 entry_point.add_command(churn.main, "churn")
 entry_point.add_command(clientsdaily.main, "clients_daily")
 entry_point.add_command(experimentsdaily.main, "experiments_daily")
+entry_point.add_command(retention.main, "retention")
```
jezdez: So this is a great example where the magic star import in the `mozetl.engagement.retention` is confusing. Just import the main function from the job module instead of relying on the start import. So this should be either:

```python
from mozetl.engagement.churn import job as churn_job
from mozetl.engagement.retention import job as retention_job
# ..
entry_point.add_command(churn_job.main, "churn")
# ..
entry_point.add_command(retention_job.main, "retention")
```

Or explicitly import the main function (and whatever you deem a callable that you want to expose in the new packages) in `mozetl/engagement/retention/__init__.py`:

```python
from .job import main  # noqa
```

and `mozetl/engagement/retention/__init__.py`:

```python
from .job import main  # noqa
```

I personally like the first better since for a submodule several levels deeps it's kind of easy to forget to maintain the API of the callables that exposed as API there.

mozetl/engagement/churn/job.py
```diff
@@ -33,8 +33,8 @@
 from pyspark.sql import SparkSession, functions as F
 from pyspark.sql.window import Window
 
-from mozetl.churn import utils, release
-from mozetl.churn.schema import churn_schema
+from mozetl.engagement.churn import release, utils
```
jezdez: Just use `from . import release, utils`

mozetl/engagement/retention/job.py
```diff
@@ -0,0 +1,166 @@
...
+
+    return cleaned_data
+
+
+def save(cleaned, path):
+    (
+        cleaned
+        .write
+        .parquet(path, mode="overwrite")
+    )
```
jezdez: What happened here? This can be:
```python
cleaned.write.parquet(path, mode="overwrite")
```

mozetl/engagement/retention/job.py
```diff
@@ -0,0 +1,166 @@
...
+         period, slack, sample):
+    """Compute churn / retention information for unique segments of
+    Firefox users acquired during a specific period of time.
+    """
+    spark = (
+        SparkSession
+        .builder
+        .appName("retention")
+        .getOrCreate()
+    )
```
jezdez: This formatting is a bit odd. I suggest to try to follow PEP8, e.g.:

```python
spark = SparkSession.builder.appName("retention").getOrCreate()
```

tests/engagement/conftest.py
```diff
@@ -1,5 +1,28 @@
+import functools
+
 import pytest
-from mozetl.churn import release
+
+from data import main_summary_sample, main_summary_schema, new_profile_sample, \
+    new_profile_schema
```
jezdez: No backslashes in imports please. It's strongly suggested in PEP8 to use Python's line continuation in parentheses, and also use absolute imports:
```
from .data import (
    main_summary_sample, main_summary_schema, new_profile_sample
    new_profile_schema
)
```

Or alternatively import data and use it as a namespace in the code:

```python
from . import data
# ..
print(data.new_profile_schema)
```

tests/engagement/test_daily_retention.py
```diff
@@ -0,0 +1,65 @@
+import arrow
+import pytest
+from pyspark.sql import functions as F
+
+from data import WEEK_START_DS, SUBSESSION_START, format_ssd, format_pcd
```
jezdez: Absolute imports please.

mozetl/engagement/retention/job.py
```diff
@@ -0,0 +1,166 @@
...
+    ]
+    return (
+        F.when(reduce(operator.__and__, is_valid_pcd), pcd)
+        .otherwise(F.lit(None))
+    )
+
+
+def transform(main_summary, start_ds):
+
+    client_date = utils.to_datetime('subsession_start_date', "yyyy-MM-dd")
```
jezdez: Why the mixed quotes here, just use single quotes.

mozetl/engagement/retention/job.py
```diff
@@ -0,0 +1,166 @@
...
+    spark = (
+        SparkSession
+        .builder
+        .appName("retention")
+        .getOrCreate()
+    )
+    spark.conf.set("spark.sql.session.timeZone", "UTC")
+
+    # If this job is scheduled, we need the input date to lag a total of
+    # 10 days of slack for incoming data. Airflow subtracts 7 days to
```
mreid-moz: nit: `slack` is no longer fixed at 10 days, and this is no longer a weekly report/job.

mozetl/engagement/churn/__init__.py
acmiyaguchi: :+1:

The initial idea was to be able to import the main entry point from churn and retention in the engagement module without having to snake through the entire project to get to it.

Originally, this used to be called as `from mozetl.churn import churn; churn.main`. It then got refactored into an 'engagement/churn/main.py` file. I thought the repetition of the module and script name was redundant and tried to import the entry point into module scope. I hadn't figured out what to call the file with the `extract`, `transform`, and `save`/`load` functions yet, so I left it as `main.py`.

I did happen to reuse `extract` in daily retention, so I imported it into module scope along with everything else in `main.py`. In retrospect, muddying up the namespace doesn't seem like a worthwhile trade-off.

It would be nice to export a stable api here (but I don't have a concrete idea on what that would look like).

mozetl/cli.py
acmiyaguchi: I went ahead with the first suggestion, I think I want to be explicit and straightforward with these changes.

mozetl/engagement/churn/job.py
```diff
@@ -33,8 +33,8 @@
...
+from mozetl.engagement.churn.schema import churn_schema
 from utils import (
```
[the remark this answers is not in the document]
acmiyaguchi: This is an artifact from refactoring, these were originally part of churn and then later added to utils. I'm going to import the variables into local scope and leave the functions as part of the utility namespace.

mozetl/engagement/retention/job.py
acmiyaguchi: I've seen this style used in Spark, [particularly in scala](https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/SparkSessionBuilderSuite.scala#L39-L43). I guess it makes sense to keep things on a single line since we read left to right.

mozetl/engagement/retention/job.py
acmiyaguchi: When you realize parentheses can be used to format chained method calls, everything looks like a nail.

acmiyaguchi:
@jezdez Thank you for the feedback! I try to keep PEP 8 in mind when I'm developing, but sometimes I write in a peculiar style. Overall, I think it would be nice to see mozetl become more consistent across the board. There are a few linter rules that I'd like to add eventually, e.g. dealing with the long method chaining as a result of the `pyspark.sql` api. It would also be nice to have some common scaffolding and structure for building scripts across our toolchain, behind the command-line api that's used to talk to Airflow. 

@mreid-moz RFAL

mozetl/engagement/retention/job.py
```diff
@@ -0,0 +1,191 @@
...
+
+    return cleaned_data
+
+
+def save(cleaned, path):
+    cleaned.write.parquet(path, mode='overwrite')
+
+
+@click.command()
+@click.option('--start_date', required=True)
```
mreid-moz: nit: convention seems to be `--start-date`

mozetl/engagement/retention/job.py
```diff
@@ -0,0 +1,191 @@
...
+@click.option('--path', default='retention_intermediate',
+              help='Path on hdfs to store intermediate data')
```
mreid-moz: Are we actually storing data on HDFS?

acmiyaguchi: Writing to "local disk" in spark on EMR ends up writing to HDFS. I'll reword the message though.

mreid-moz: I thought HDFS on EMR was effectively ephemeral - don't we want to save to S3 for "persistent" storage?

acmiyaguchi: The data is local to the cluster so that I don't have to worry about clean-up. The intermediate data isn't useful other than for aggregates (and it can always be generated again via this script),

mreid-moz: Ahh, gotcha. Makes sense now, thanks!

PR #135, opened 2017-10-05
Add shutdown_kill aggregate

mreid-moz:
Use a `sum` aggregation


acmiyaguchi:
Looks good. What's the context for this patch?

mreid-moz:
Saptarshi was hoping to use it as part of measuring crashiness. It seemed clear that the right aggregation was "sum", so no reason not to add it.

PR #152, opened 2017-10-10
Pin and sort python dependencies

wlach:


setup.py
```diff
@@ -24,14 +24,14 @@
...
-        'click',
-        'requests',
-        'click_datetime',
-        'arrow',
-        'scipy',
-        'numpy'
+        'arrow==0.10.0',
+        'click==6.7',
+        'click_datetime==0.2',
+        'numpy==1.13.3'
```
acmiyaguchi: There's a missing comma here.

wlach:
Eek, not sure how I missed the comma issue.

I think for a package like this the best strategy is to only upgrade dependencies if there is some new feature or fix that you want to take advantage of. The list is small enough that you can probably just do this manually.

setup.py
```diff
@@ -24,14 +24,14 @@
...
-        'requests',
-        'click_datetime',
-        'arrow',
-        'scipy',
-        'numpy'
+        'arrow==0.10.0',
+        'click==6.7',
+        'click_datetime==0.2',
+        'numpy==1.13.3',
+        'pyspark==2.2.0',
```
acmiyaguchi: I've tested this locally using `tox -r`, and it seems like I get the same error. Running `pip freeze | grep pyspark` gives `pyspark==2.2.0`, but tox is expecting `2.2.0.post0`. Possibly a packaging error on Spark's side?

setup.py
```diff
@@ -24,14 +24,14 @@
...
+        'python_moztelemetry==0.9.8',
```
acmiyaguchi: According to pip it should be `0.8.9` instead of `0.9.8`.

wlach:
Yeah, I noticed the build was failing as well, mean to fix it and then ask for re-review, but you beat me to it. :)

wlach:
@acmiyaguchi ok I think we're good now. travis is happy

PR #153, opened 2017-10-10
Add legacy addons etl job to TAAR

mlopatka:
Just a small job that uses a new AMO API to get appropriate replacement addons for legacy addons. This will populate a data source for one of TAAR's modules.

### DIFF
--- a/mozetl/taar/taar_legacy.py
+++ b/mozetl/taar/taar_legacy.py
@@ -1,0 +1,59 @@
+"""
+TAAR legacy addon replacements
+This job pulls the latest curated substitute webestensions from the
+AMO data sources. This ensures that TAAR recommendations for legacy
+substitutions are consistent with AMO replacements.
+Prototype notebook is here [1]
+
+[1] https://gist.github.com/mlopatka/c6fe350dbcd8fe11ccf0929fc9f8c223
+"""
+
+import click
+import json
+import logging
+import requests
+
+from taar_utils import store_json_to_s3
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+AMO_LEGACY_RECS_URI =\
+    "https://addons.mozilla.org/api/v3/addons/replacement-addon/"
+EXPORT_FILE_NAME = 'legacy_substitutes.json'
+
+
+def fetch_legacy_replacement_masterlist():
+    # Set the initial URI to the REST API entry point.
+    request_uri = AMO_LEGACY_RECS_URI
+
+    # Put all the mappings in this dictionary, keyed on legacy guid.
+    legacy_replacement_dict = {}
+
+    while request_uri is not None:
+        print("Fetching {}\n".format(request_uri))
+        # Request the data and get it in JSON format.
+        r = requests.get(request_uri)
+        page_blob = r.json()
+        # Re map as {guid : [guid-1, guid-2,... guid-n]}
+        for addon in page_blob.get('results'):
+            legacy_replacement_dict[addon.get('guid')] =\
+                addon.get('replacement')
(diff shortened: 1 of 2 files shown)


Dexterp37:
Anthony, do you have any cycles to review this?

mozetl/taar/taar_legacy.py
```diff
@@ -0,0 +1,59 @@
...
+
+@click.command()
+@click.option('--date', required=True)
+@click.option('--bucket', default='telemetry-private-analysis-2')
+@click.option('--prefix', default='taar/legacy/')
+def main(date, bucket, prefix):
+    logger.info("Retreiving AMO legacy addon replacements list")
+    legacy_dict = fetch_legacy_replacement_masterlist()
+
+    if len(legacy_dict) > 0:
```
acmiyaguchi: You might consider adding a logging statement if the legacy dictionary is empty, for diagnostic purposes.

mozetl/taar/taar_legacy.py
```diff
@@ -0,0 +1,59 @@
...
+
+def fetch_legacy_replacement_masterlist():
+    # Set the initial URI to the REST API entry point.
+    request_uri = AMO_LEGACY_RECS_URI
+
+    # Put all the mappings in this dictionary, keyed on legacy guid.
+    legacy_replacement_dict = {}
+
+    while request_uri is not None:
+        print("Fetching {}\n".format(request_uri))
```
acmiyaguchi: Use the logger to write to stdout.

tests/test_taar_legacy.py
```diff
@@ -0,0 +1,18 @@
...
+
+import pytest
+import requests
+
+AMO_LEGACY_RECS_URI =\
+    "https://addons-dev.allizom.org/api/v3/addons/replacement-addon/"
+
+
+@pytest.fixture
+def is_api_up():
```
acmiyaguchi: Avoid hitting a live server in the testing suite. One way to test `fetch_legacy_replacement_masterlist` would be to mock/monkeypatch `requests.get` with a custom function that reads from a predefined dictionary. You can specify in the test what the schema of the data you're expecting looks like and double check that the execution of the main routine is working as expected.

The doc page for the [pytest monkeypatch fixture](https://docs.pytest.org/en/latest/monkeypatch.html) is pretty good.  Here's [an example](https://github.com/mozilla/python_mozetl/blob/master/tests/test_topline_summary.py#L316-L318) of monkeypatch being used in this repository for tests.

mlopatka: Awesome! Thanks for the detailed feedback. PR has been updated to include appropriate test coverage in this manner.

mozetl/taar/taar_legacy.py
mlopatka: made these changes on the line below in an ``else'' condition.

tests/test_taar_legacy.py
```diff
@@ -0,0 +1,84 @@
...
+    "guid-26": [],
+    "guid-27": {},
+    "guid-28": None,
+    "guid-29": ["guid-30"]
+}
+
+
+@pytest.fixture
+def mock_amo_api_response(monkeypatch):
+    monkeypatch.setattr(requests, 'requests.get', lambda x, y: FAKE_LEGACY_DATA)
```
acmiyaguchi: On a second look, it seems that the second argument has an extraneous `requests.` prefixed since the module was already specified in the first argument.

mlopatka:
Hold off on the PR, I have some compatibility issues between my local environment and the project requirements. I'm getting inconsistent behavior. I'll submit a PR with squashed commits once this is resolved.

PR #155, opened 2017-10-12
Add taar legacy job with test coverage

mlopatka:


mlopatka:
@acmiyaguchi this is the replacement for PR#154. I resolved my local environment issues and it should pass tests now.

PR #157, opened 2017-10-19
Bug 1409219 - snakecase columns when using sync_summary_v2

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1409219)

@kitcambridge r?

### DIFF
--- a/mozetl/sync/bookmark_validation.py
+++ b/mozetl/sync/bookmark_validation.py
@@ -37,14 +37,14 @@ def transform(spark):
            s.app_name,
            s.app_channel,
            s.uid,
-           s.deviceID AS device_id,
+           s.device_id AS device_id,
            s.submission_date_s3 AS submission_day,
            date_format(from_unixtime(s.when / 1000), 'YYYYMMdd') AS sync_day,
            s.when,
            s.status,
            e.name AS engine_name,
            e.status AS engine_status,
-           e.failureReason AS engine_failure_reason,
+           e.failure_reason AS engine_failure_reason,
            e.validation.problems IS NOT NULL AS engine_has_problems,
            e.validation.version AS engine_validation_version,
            e.validation.checked AS engine_validation_checked,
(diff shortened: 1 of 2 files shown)


mozetl/sync/bookmark_validation.py
```diff
@@ -105,10 +105,10 @@ def load(spark, bucket, prefix, version, start_date):
 @click.command()
 @click.option('--start_date', required=True, help="Date to process")
 @click.option('--end_date', help="Optional end date to run until")
-@click.option('--bucket', default='net-mozaws-prod-us-west-2-pipeline-analysis')
-@click.option('--prefix', default='kit/sync')
+@click.option('--bucket', default='telemetry-parquet')
+@click.option('--prefix', default='sync')
 @click.option('--input_bucket', default='telemetry-parquet')
-@click.option('--input_prefix', default='sync_summary/v1')
+@click.option('--input_prefix', default='sync_summary/v2')
```
kitcambridge: Do we need to set `.option("mergeSchema", "true")` in `extract`, when we call `spark.read`, or will this do the right thing when backfilling older data (or do we not need to worry about that)?

acmiyaguchi: The dataset isn't backward compatible because the column names don't match. You would have to use the script from the old revision to backfill everything before 2017-05-05(?).

tests/test_sync_bookmark.py
```diff
@@ -0,0 +1,320 @@
...
+                        ]
+                    }
+                }
+            ]
+        },
+    ])
+
+    assert df.count() == 4
+
+    # one hot encode the cases
```
mreid-moz: nit: I don't understand this comment

acmiyaguchi: You can tell which case failed based on the digit. Case 1 corresponds to digit 1 (0001), case 3 corresponds to digit 3 (0100). I still had to drop into pdb to debug the cases though.

PR #160, opened 2017-10-23
 Bug 1404133 - Refactor the search_aggregates job so it's easy to run from a notebook

wlach:


wlach:
just tested this inside a notebook and it seems to work pretty well!

PR #163, opened 2017-10-25
Bug 1394610 - Add the follow-on search addon version to the search_dashboard dataset

wlach:


wlach:
This is a little slower than I would like: about 740 seconds on a single node cluster on a 1% sample of main_summary. I tried a more efficient strategy for getting the data out of the addons structure in the udf, but it only improved things by a small amount (10 seconds).

harterrt:
Eh, that speed doesn't set off alarms for me. It would be nice to be faster, but we can backfill a year of data in ~50 machine hours. Easily done in a day with a 10 node cluster.

harterrt:
That said, feel free to continue optimizing if you'd like. I'd love to see your findings.

wlach:
It is hard for me to restrain myself from optimizing something that I feel should be faster, but you're almost certainly right that this is fast enough. 😄 

I suspect creating a temporary table with the exploded addons list, filtering it to only include the search follow on, then joining that with the existing table would be faster... but that can wait until later.

PR #167, opened 2017-10-26
Log how long it took to produce a day's worth of search aggregate rollups

wlach:


mreid-moz:
@wlach is it ok to merge this?

wlach:
Yup! I will do so now.

PR #168, opened 2017-10-26
Add "mean" for unique domains

mreid-moz:
Request from :joy on IRC.

Also clean up the tests a bit.

### DIFF
--- a/tests/test_clientsdaily.py
+++ b/tests/test_clientsdaily.py
@@ -1,6 +1,7 @@
 import pytest
 import os
 from mozetl.schemas import MAIN_SUMMARY_SCHEMA
+from mozetl.clientsdaily import rollup as cd
 
 
 EXPECTED_INTEGER_VALUES = {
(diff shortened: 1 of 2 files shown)


tests/test_clientsdaily.py
```diff
@@ -22,27 +25,29 @@ def make_frame(spark):
...
+def clients_daily(main_summary_with_search):
+    return cd.to_profile_day_aggregates(main_summary_with_search)
+
 
-    frame = make_frame(spark)
-    extracted = rollup.extract_search_counts(frame)
-    row = extracted.agg({'search_count_all': 'sum'}).collect()[0]
+def test_extract_search_counts(main_summary_with_search):
+    row = main_summary_with_search.agg({'search_count_all': 'sum'}).collect()[0]
     total = row.asDict().values()[0]
```
acmiyaguchi: This could also be done with `df.first()[0]`, where `.first()` returns a `Row`, and `Row[0]` returns the first item.

PR #171, opened 2017-11-10
Bug 1416225 - skip nulls with "first" aggregation in clients_daily

mreid-moz:
I can't imagine a case where a null would be preferred over the first non-null value, particularly given that the data is not sorted first, as far as I can tell.

### DIFF
--- a/tests/test_clientsdaily.py
+++ b/tests/test_clientsdaily.py
@@ -117,3 +117,19 @@ def test_sessions_started_on_this_day_sorted(clients_daily):
     ten_ssotds = one_day.select("sessions_started_on_this_day").take(10)
     actual = [r.asDict().values()[0] for r in ten_ssotds]
     assert actual == expected
+
+
+# Ensure that "first" aggregations skip null values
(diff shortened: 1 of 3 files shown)


tests/test_clientsdaily.py
```diff
@@ -117,3 +117,19 @@ def test_sessions_started_on_this_day_sorted(clients_daily):
...
     actual = [r.asDict().values()[0] for r in ten_ssotds]
     assert actual == expected
+
+
+# Ensure that "first" aggregations skip null values
+def test_first_skips_nulls(clients_daily):
+    filter_template = "client_id = '{}' and activity_date = '{}'"
+    client = '0c495fce-5fbf-4f4a-ac03-2dedcef0a8d0'
+    day = '2017-05-25'
+    filter_clause = filter_template.format(client, day)
```
acmiyaguchi: I assume this case is where `sync_configured` is `null`.

mreid-moz: yup, that was a case where the first entry was null, and a later entry was non-null.

PR #173, opened 2017-11-15
Bug 1415642 - Make experiments daily incremental

relud:


relud:
sure. should that go in [mozetl/utils.py](https://github.com/mozilla/python_mozetl/blob/master/mozetl/utils.py)?

mreid-moz:
yep, that sounds good.

mozetl/utils.py
```diff
@@ -131,3 +131,35 @@ def send_ses(fromaddr,
...
+    Note that the start_date will be `lag-days` days before date. In other
+    words, if you pass in 2017-01-20 and set `lag-days` to 5, the aggregation
+    will be processed for day 2017-01-15 (the resulting data will cover
+    submission dates including the activity day itself plus 5 days of lag for
+    a total of 6 days).
+    """
+    end_date = DT.datetime.strptime(date, '%Y-%m-%d').date()
+    # Rewind by `lag_days` to the desired activity date.
+    start_date = end_date - DT.timedelta(lag_days)
+    ACTIVITY_DATE_COLUMN = F.expr(
```
mreid-moz: It doesn't look like `F` is imported in this file, so this may not work.

I think it would be preferable to import `ACTIVITY_DATE_COLUMN` from `fields.py` to keep a single definition for this.

relud: fixed

mozetl/experimentsdaily/rollup.py
```diff
@@ -46,21 +50,34 @@ def to_experiment_profile_day_aggregates(frame_with_extracts):
 @click.option('--output-version',
               default=2,
```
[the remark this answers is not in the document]
relud: fixed

PR #174, opened 2017-11-28
Bug 1419080 - Use attribution from new-profile pings in churn 

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1419080)

Attribution is [often] missing from main-pings. This uses the attribution value from the new-profile ping over the main-ping value when generating churn and retention.

The left join will most likely increase the processing time on this dataset.

### DIFF
--- a/mozetl/engagement/churn/job.py
+++ b/mozetl/engagement/churn/job.py
@@ -508,6 +508,12 @@ def main(start_date, bucket, prefix, input_bucket, input_prefix,
     )
     spark.conf.set("spark.sql.session.timeZone", "UTC")
 
+    shuffle_partitions = max(
+        spark.sparkContext.defaultParallelism * 4,
+        int(spark.conf.get("spark.sql.shuffle.partitions"))
+    )
+    spark.conf.set("spark.sql.shuffle.partitions", shuffle_partitions)
+
     # If this job is scheduled, we need the input date to lag a total of
     # 10 days of slack for incoming data. Airflow subtracts 7 days to
     # account for the weekly nature of this report.
--- a/mozetl/engagement/churn/job.py
+++ b/mozetl/engagement/churn/job.py
@@ -114,6 +114,33 @@ def clean_new_profile(new_profile):
     return new_profile.select(utils.build_col_expr(select_expr))
 
 
+def join_and_coalesce(lhs, rhs, on, col):
+    """ Join two dataframes on a common column, coalescing the value of a
+    target column to the right hand side over the left hand side.
+
+    :param lhs: dataframe as the frame of reference, left hand side
+    :param rhs: dataframe to coalesce values from, right hand side
+    :param on:  column name to join on
+    :param col: name of the column in the rhs to coalesce with lhs
+    :return:    a dataframe with the same schema as the lhs
+    """
+
+    assert len({on, col} & set(lhs.columns) & set(rhs.columns)) == 2, \
+        "both dataframes should contain join key and coalesce column"
+
+    rhs_name = "_rhs_column_name"
+    rhs = rhs.select(on, F.col(col).alias(rhs_name))
+
+    coalesced = (
+        lhs
+        .join(rhs, on, "left")
+        .withColumn(col, F.coalesce(rhs_name, col))
+        .select(lhs.columns)
+    )
+
+    return coalesced
+
+
 def extract(main_summary, new_profile, start_ds, period, slack, is_sampled):
     """
     Extract data from the main summary table taking into account the
@@ -139,15 +166,23 @@ def extract(main_summary, new_profile, start_ds, period, slack, is_sampled):
     if is_sampled:
         predicates.append((F.col("sample_id") == "57"))
 
-    return (
+    extract_ms = (
         main_summary
         .where(reduce(operator.__and__, predicates))
         .select(SOURCE_COLUMNS)
-        .union(
-            clean_new_profile(new_profile)
-            .where(reduce(operator.__and__, predicates))
-            .select(SOURCE_COLUMNS)
-        )
+    )
+
+    np = clean_new_profile(new_profile)
+    extract_np = (
+        np
+        .where(reduce(operator.__and__, predicates))
+        .select(SOURCE_COLUMNS)
+    )
+
+    return (
+        # bug 1416364: use attribution from new profiles
+        join_and_coalesce(extract_ms, np, "client_id", "attribution")
+        .union(extract_np)
     )
 
 
--- a/mozetl/engagement/churn/job.py
+++ b/mozetl/engagement/churn/job.py
@@ -388,7 +388,15 @@ def clean_columns(prepared_clients, effective_version, start_ds):
             (F.col("profile_creation") > DEFAULT_DATE) &
             (pcd <= client_date)
         ))
-        .select("profile_creation", *select_expr)
+        .select(
+            # avoid acquisition dates in the future
+            (
+                F.when(F.col(is_valid), F.col("profile_creation"))
+                .otherwise(F.lit(None))
+                .alias("profile_creation")
+            ),
+            *select_expr
+        )
         # Set default values for the rows
         .fillna({
             'acquisition_period': DEFAULT_DATE,
--- a/mozetl/engagement/churn/job.py
+++ b/mozetl/engagement/churn/job.py
@@ -179,9 +179,16 @@ def extract(main_summary, new_profile, start_ds, period, slack, is_sampled):
         .select(SOURCE_COLUMNS)
     )
 
+    np_attribution = (
+        np
+        .where("attribution is not null")
+        .groupBy("client_id")
+        .agg(F.first("attribution").alias("attribution"))
+    )
+
     return (
         # bug 1416364: use attribution from new profiles
-        join_and_coalesce(extract_ms, np, "client_id", "attribution")
+        join_and_coalesce(extract_ms, np_attribution, "client_id", "attribution")
         .union(extract_np)
     )
 
--- a/mozetl/engagement/churn/job.py
+++ b/mozetl/engagement/churn/job.py
@@ -114,31 +114,38 @@ def clean_new_profile(new_profile):
     return new_profile.select(utils.build_col_expr(select_expr))
 
 
-def join_and_coalesce(lhs, rhs, on, col):
-    """ Join two dataframes on a common column, coalescing the value of a
-    target column to the right hand side over the left hand side.
+def coalesce_new_profile_attribution(main_summary, new_profile):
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
Fails with the following error:

`Caused by: org.apache.spark.SparkException: Job aborted due to stage failure: ShuffleMapStage 11 (parquet at NativeMethodAccessorImpl.java:0) 
has failed the maximum allowable number of times: 4. Most recent failure reason: org.apache.spark.shuffle.FetchFailedException: Too large fram
e: 6374624378 `

I think I might need to
* increase the number of nodes
* repartition before joining against *all* of the new-profiles

acmiyaguchi:
I profiled two approaches in terms of improving the join performance on a standalone databricks cluster (c3.4xlarge x 10). 

* ms left join np
* ms left join (np right join ms)

The following query corresponds to the first case and runs in 11.07 minutes.
```sql
with ms as (
  select *
  from main_summary
  where submission_date_s3 >= '20171112' and submission_date_s3 <= '20171122'
),
np as (
  select client_id, environment.settings.attribution
  from telemetry_new_profile_parquet
  where environment.settings.attribution is not null
),
coalesced as (
  select ms.client_id, coalesce(np.attribution, ms.attribution) as attribution
  from ms
  left join np on ms.client_id = np.client_id
)

select attribution, count(distinct client_id) from coalesced where group by 1 order by 1
```
The second case handles the increasing size of the new profile dataset. This query runs in 14.88 minutes and is just as correct. I don't think it's necessary to modify the current job yet. 

The relevant portion of the query follows:
```sql
np as (
  select np.client_id, environment.settings.attribution
  from telemetry_new_profile_parquet np
  right join (
    select distinct client_id
    from ms
  ) ms on ms.client_id = np.client_id 
  where environment.settings.attribution is not null
),
```

I don't think there are too many ways to improve the performance of the join and coalesce here, so I'll just bump up the size of the cluster once I verify that this patch works correctly in ATMO. 

[link to notebook with results](https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/3507/command/3513)

fbertsch:
@acmiyaguchi a few points:
- It seems your two queries are doing the same thing, the second one will just take longer (as you noticed)
- I would hazard that your issue comes from a few clients that have both many profiles and many pings. See [0] and [1], which shows a single client that will have over 5.5 Billion rows if joined to `main_summary` on `client_id` (for two months of `main_summary` - if you're using ten days this will be obviously less). All of these pings will end up on the same partition during the join.
- Also consider bumping up `spark.sql.shuffle.partitions`, which is set to 200 by default.

acmiyaguchi:
@frank It slipped my mind that there would be more than one new profile ping per client_id, it is definitely increasing the size of the dataset significantly.

acmiyaguchi:
@fbertsch this is ready for review, I've verified that it no longer crashes on 5 nodes (but takes 4 hours). I'll need to bump the number of workers up to compensate.

mozetl/engagement/churn/job.py
```diff
@@ -139,15 +170,30 @@ def extract(main_summary, new_profile, start_ds, period, slack, is_sampled):
...
+        np
+        .where(reduce(operator.__and__, predicates))
+        .select(SOURCE_COLUMNS)
+    )
+
+    np_attribution = (
+        np
+        .where("attribution is not null")
+        .groupBy("client_id")
+        .agg(F.first("attribution").alias("attribution"))
```
fbertsch: Can you confirm that `attribution` is stable across `new_profile` pings?

mozetl/engagement/churn/job.py
```diff
@@ -139,15 +170,30 @@ def extract(main_summary, new_profile, start_ds, period, slack, is_sampled):
...
+    np = clean_new_profile(new_profile)
+    extract_np = (
+        np
+        .where(reduce(operator.__and__, predicates))
+        .select(SOURCE_COLUMNS)
+    )
+
+    np_attribution = (
+        np
+        .where("attribution is not null")
```
fbertsch: This filter is now done twice, here and in line 134.

tests/engagement/test_churn.py
```diff
@@ -383,3 +384,68 @@ def case(name, expect, desktop, mobile, configured):
...
+        env['settings']['attribution'] = attribution
+        return env
+
+    # The case for NULL in new_profile and NOT NULL in main_summary is
+    # explicitly not tested here because it is not a well defined behavior.
+    new_profile = generate_new_profile_data([
+        {'client_id': '2', 'environment': update_attribution({'source': 'mozilla.org'})},
+        # overwrite the null value
+        {'client_id': '3', 'environment': update_attribution({'source': 'mozilla.org'})},
+        # overwrite the null value, but is not a ping of itself
```
fbertsch: What does this mean?

tests/engagement/test_churn.py
```diff
@@ -383,3 +384,68 @@ def case(name, expect, desktop, mobile, configured):
...
+        {
+            'client_id': '4',
+            'environment': update_attribution({'source': 'mozilla.org'}),
+            'submission': SUBSESSION_START.shift(days=-7).format("YYYYMMDD"),
+        },
+        # not counted toward total
```
fbertsch: Please describe why it's not counted.

mozetl/engagement/churn/job.py
acmiyaguchi: It looks like there can be [multiple attribution codes per client](https://sql.telemetry.mozilla.org/queries/49745) in the new-profile pings. It looks like an overwhelming number of clients [appear once with a single configuration](https://sql.telemetry.mozilla.org/queries/49746/source#133842). I don't think there's a good way of getting around this without fixing attribution at the client level.

mozetl/engagement/churn/job.py
acmiyaguchi: I'm removing this from line 134, it makes sense to push down the filter.

mozetl/engagement/churn/job.py
fbertsch: Okay, it seems fine to me to proceed by using `first` in this case. Go ahead and publish that attribution query, it seems useful to have around.

tests/engagement/test_churn.py
acmiyaguchi: Poorly written, it's the case where the new-profile is used to recover attribution but is outside of the current retention period. I make this distinction because I use new-profile pings to recover clients that don't appear again.

tests/engagement/test_churn.py
```diff
@@ -383,3 +384,68 @@ def case(name, expect, desktop, mobile, configured):
...
+    ])
+
+    def update_attribution(attribution):
+        # only updates the attribution section in the environment
+        env = copy.deepcopy(data.new_profile_sample['environment'])
+        env['settings']['attribution'] = attribution
+        return env
+
+    # The case for NULL in new_profile and NOT NULL in main_summary is
+    # explicitly not tested here because it is not a well defined behavior.
```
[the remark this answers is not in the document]
acmiyaguchi: I'd like to leave the case documented, but I don't think breaking the case is necessarily a bad thing.

In order to avoid double counting clients, the attributes from the first ping of a client taken as a representative.

Essentially

``` sql
select
    attribution, 
    count(client_id) as n_profiles
from (
    select
        client_id,
        attribution,
        row_number() over (order by timestamp desc) as rank
    from main_summary
)
where rank = 1
group by 1
```
as opposed to
```sql
select
    attribution,
    count(distinct client_id) as n_profiles
from main_summary
group by 1
```

Actually after writing that out, I guess it's a good idea to document the quirk in the test. The second query is only viable when you can avoid double counting (i.e. HLL variable).

acmiyaguchi:
@fbertsch ready for another look. I addressed most of the review points, the test should be cover most of the important cases.

tests/engagement/test_churn.py
```diff
@@ -383,3 +384,87 @@ def case(name, expect, desktop, mobile, configured):
...
+        # recover null attribution
+        {'client_id': '3', 'environment': update_attribution({'source': 'mozilla.org'})},
+        # new-profile ping used to recover attribution, but outside of the
+        # the current retention period
+        {
+            'client_id': '4',
+            'environment': update_attribution({'source': 'mozilla.org'}),
+            'submission': SUBSESSION_START.shift(days=-7).format("YYYYMMDD"),
+        },
+        # ping outside of the current retention period not counted toward totals
```
fbertsch: This isn't necessarily being tested here, since `5` is not in the main_summary dataset (it's testing what `2` is testing: a new profile without a main summary companion).

mozetl/engagement/churn/job.py
```diff
@@ -114,6 +114,33 @@ def clean_new_profile(new_profile):
     return new_profile.select(utils.build_col_expr(select_expr))
 
 
+def join_and_coalesce(lhs, rhs, on, col):
```
fbertsch: I do think that this function now needlessly complicates the code. Up to you to leave it in.

tests/engagement/test_churn.py
```diff
@@ -383,3 +384,87 @@ def case(name, expect, desktop, mobile, configured):
...
+            "client_id": "created in the future",
+            "profile_creation_date": data.format_pcd(data.SUBSESSION_START.shift(days=+10))
+        },
+        {"client_id": "sanity"},
+    ])
+
+    # null and the single sane row
+    assert df.count() == 2
+
+    # attributes head to null space if pcd is invalid
```
fbertsch: Can you elaborate on why the above test has anything to do with `geo`?

acmiyaguchi:
I've added refactored the function a bit and changed the 5th test case to cover the case where `attribution = {}`.

[This comment in the related bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1416364#c4) is causing me to reconsider this PR.

* this PR recovers roughly 15% of attributed profiles from the week of 20171112
* recovering attribution increases the compute time significantly and introduces subtle behavior wrt multi-attribution clients
* new-profiles after buildid 20170924 should set attribution correctly

This is still probably worthwhile to maintain and the code is self contained if it needs to be removed later.

mozetl/engagement/churn/job.py
```diff
@@ -114,6 +114,39 @@ def clean_new_profile(new_profile):
...
+        "attribution.medium",
+        "attribution.campaign",
+        "attribution.content",
+    ]
+
+    # all pings in new-profile that contain attribution
+    np_attribution = (
+        new_profile
+        .where(F.col("attribution").isNotNull() &
+               F.coalesce(*nested_attr_cols).isNotNull())
```
fbertsch: Please add a comment here on why this is now required.

PR #178, opened 2018-01-12
Don't include clients with absurd search counts in aggregation

wlach:


wlach:
This hopefully fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1412040

mozetl/search/dashboard.py
```diff
@@ -70,6 +70,7 @@ def explode_search_counts(main_summary):
         .withColumn(exploded_col_name, explode(col('search_counts')))
         .withColumn('addon_version', udf_get_search_addon_version('active_addons'))
         .select(['*'] + search_fields)
+        .filter('single_search_count.count < 1000')
```
harterrt: Can we bump this up to 10k to be safe? IIRC, the outrageous values tend to be hundreds of millions of searches per day.

wlach: Makes sense, I'll do that

fbertsch: Drive-by comment: move this to a well-named constant which can also be used in the tests.

wlach: Also a good idea, done

PR #180, opened 2018-01-18
Refactor search_aggregates job to prepare for search_clients_daily

harterrt:


mozetl/basic/transform.py
```diff
@@ -40,7 +40,7 @@ def _build_cell(ping, column_config):
     if func is not None:
         try:
             return func(raw_value)
-        except:  # TODO: what exception could happen here, fill in?
+        except Exception:
```
wlach: I actually made this change in #181, we should probably merge that instead.

tests/test_search_aggregates.py
```diff
@@ -143,16 +150,16 @@ def simple_main_summary(generate_main_summary_data):
...
-        ('country',         'DE',       StringType(), True),
-        ('locale',          'de',       StringType(), True),
-        ('search_cohort',   None,       StringType(), True),
-        ('app_version',     '54.0.1',   StringType(), True),
-        ('distribution_id', None,       StringType(), True),
-        ('addon_version',   '0.9.5',    StringType(), False),
-        ('engine',          'google',   StringType(), False),
-        ('source',          'urlbar',   StringType(), False),
-        ('count',           4,          LongType(),   False),
+        ('submission_date', '20170101',    StringType(), False),
```
wlach: I think I would prefer if we didn't pad the spacing in these structures as it makes diffs harder to read. Also, I don't use an editor that makes it easy to add the appropriate level of padding.

mozetl/search/aggregates.py
```diff
@@ -125,8 +159,9 @@ def generate_dashboard(submission_date, bucket, prefix,
     main_summary = spark.read.parquet(source_path)
 
     logger.info('Running the search dashboard ETL job...')
-    search_dashboard_data = search_dashboard_etl(main_summary)
+    search_dashboard_data = search_aggregates(main_summary)
 
+    # TODO: move this to a function and simplify saving
```
wlach: Not sure about this comment, I don't think things are so bad the way they are?

mozetl/basic/transform.py
harterrt: SGTM, I'll remove this change.

tests/test_search_aggregates.py
harterrt: I'm starting to agree with you. The upcoming search_clients_daily PR has some tests that break this padding. I'll remove the padding generally in that PR. SG?

PR #181, opened 2018-01-19
Fix use of bare except in basic transforms

wlach:
Bare except makes it harder to interrupt programs. Better to catch the Exception
class instead, which will at least let the SystemExit and KeyboardInterrupt
exceptions go through (https://stackoverflow.com/a/14797463).

### DIFF
--- a/mozetl/basic/transform.py
+++ b/mozetl/basic/transform.py
@@ -1,7 +1,12 @@
+import logging
 from collections import namedtuple
 
 from moztelemetry import get_pings_properties
 from pyspark.sql.types import StructType, StructField
+
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
 
 
 ColumnConfig = namedtuple('ColumnConfig',
@@ -40,7 +45,9 @@ def _build_cell(ping, column_config):
     if func is not None:
         try:
             return func(raw_value)
-        except Exception as e:  # TODO: should we make this more specific?
+        except Exception as e:
+            logger.warning("Unable to apply transform on data '%s': %s",
+                           raw_value, e)
             return None
     else:
         return raw_value
--- a/mozetl/basic/transform.py
+++ b/mozetl/basic/transform.py
@@ -40,7 +40,7 @@ def _build_cell(ping, column_config):
     if func is not None:
         try:
             return func(raw_value)
-        except:  # TODO: what exception could happen here, fill in?
+        except Exception as e:  # TODO: should we make this more specific?
             return None
     else:
         return raw_value


mozetl/basic/transform.py
```diff
@@ -40,7 +40,7 @@ def _build_cell(ping, column_config):
     if func is not None:
         try:
             return func(raw_value)
-        except:  # TODO: what exception could happen here, fill in?
+        except Exception as e:  # TODO: should we make this more specific?
```
jezdez: Since the "cleaning func" here can be any type of callback it's hard to know what exception we can except to be honest, which I guess is a problem of how column configs are defined. Since that's the case we can safely remove the `TODO` (or allow the cleaning function to also optionally return the exception it may raise. E.g. we could allow the `ColumnConfig.cleaning_func` attribute to return optionally a tuple with the callback and the exception. Or we could add a new attribute `ColumnConfig.cleaning_exc` which if not `None` would be used to catch issues properly. That would make this more flexible.

That said, if we allow defining a cleaning function like this (and I only found two instances [here](https://github.com/mozilla/python_mozetl/blob/bad1e83ca967213ce7c7fc5fc106800601de8d50/mozetl/testpilot/pulse.py#L82) and [here](https://github.com/mozilla/python_mozetl/blob/bad1e83ca967213ce7c7fc5fc106800601de8d50/mozetl/testpilot/pulse.py#L87)) and return `None` if they fail it's as if the data doesn't exist while it may only be wrongly formatted or a transient error etc. And doesn't that imply that we'd be hiding the fact that we found an issue with the existing data? Wouldn't it make more sense to return the raw value here instead to be able to improve the cleaning function to handle real world data?

wlach: I don't understand the context all that well tbh, but my strong feeling is that returning "None" (null) is better than trying to return unprocessed data (which might later be read as garbage) or failing altogether.

I think the best thing for now might be to remove the TODO and say exactly this. Thoughts?

jezdez: Admittedly I lack deeper understanding what the purpose or at least the basic assumptions regarding the data purity are, maybe @acmiyaguchi, @sunahsuh or @mreid-moz wants to chime in?

mreid-moz: I don't think we should just quietly hide exceptions and replace with None. At the very least, a log message should be emitted when we skip a value.

I also don't think you can safely return the raw value, since it may very well be a different type than the "cleaned" value, or might get coerced to an incorrect value.

Could we document that cleaning functions must never raise exceptions, and create light wrapper functions that handle exceptions? Then we'd just let things raise (or bail out entirely) if we encounter an exception while cleaning.
That would make the desired behaviour explicit, and customizable per cleaner.

wlach: As @jezdez pointed out there are only two instances of us using these cleaning functions, so maybe the simplest thing to do is take this try/except logic out and see what breaks. I assume that most errors that we'd see would already be caught as schema violations.

mreid-moz: Parsing `creationDate` as a date will almost certainly raise the odd exception here and there. I'm not sure about the `_requests_to_rows` function, but I would guess that replacing with `None` upon exception is probably desirable in both example cases. 

The part that concerns me is not knowing the scale or variety of failure modes.  Can we keep the existing behaviour, but add some logging so we can review the job output and see what kinds of values (if any) we've dropped due to exceptions?

PR #184, opened 2018-01-31
Bug 1431777 - add missing column for experiments daily

relud:
"scalar_parent_telemetry_os_shutting_down" isn't present in experiment
data sets, but it's needed for EXPERIMENT_FIELD_AGGREGATORS to work

This might fix experiments daily. I don't know why it worked before, or why my test runs log a traceback but real runs don't. I don't think this is actually the right place to solve this, but it is a solution, and I don't have any others yet.

### DIFF
--- a/mozetl/experimentsdaily/rollup.py
+++ b/mozetl/experimentsdaily/rollup.py
@@ -17,6 +17,19 @@ def load_experiments_summary(spark, parquet_path):
     )
 
 
+def add_missing_columns(frame):
+    """
+    to_experiment_profile_day_aggregates() needs the column
+    "scalar_parent_telemetry_os_shutting_down" for EXPERIMENT_FIELD_AGGREGATORS
+    so this function adds it if it's missing from the frame.
+    """
+    from pyspark.sql.functions import lit
+    from pyspark.sql.types import BooleanType
+    if "scalar_parent_telemetry_os_shutting_down" not in frame.columns:
+        frame = frame.withColumn("scalar_parent_telemetry_os_shutting_down", lit(None).cast(BooleanType()))
+    return frame
+
+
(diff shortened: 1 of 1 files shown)


mozetl/experimentsdaily/rollup.py
```diff
@@ -17,6 +17,19 @@ def load_experiments_summary(spark, parquet_path):
...
 
 
+def add_missing_columns(frame):
+    """
+    to_experiment_profile_day_aggregates() needs the column
+    "scalar_parent_telemetry_os_shutting_down" for EXPERIMENT_FIELD_AGGREGATORS
+    so this function adds it if it's missing from the frame.
+    """
+    from pyspark.sql.functions import lit
+    from pyspark.sql.types import BooleanType
```
relud: not sure if I should import these at the top of the file, or here, based on `to_experiment_profile_day_aggregates()`

mozetl/experimentsdaily/rollup.py
```diff
@@ -17,6 +17,19 @@ def load_experiments_summary(spark, parquet_path):
...
+    if "scalar_parent_telemetry_os_shutting_down" not in frame.columns:
+        frame = frame.withColumn("scalar_parent_telemetry_os_shutting_down", lit(None).cast(BooleanType()))
```
relud: not sure if this is the correct type for this column

relud:
also noteworthy: in my test runs experiments_daily fails without this even on 2017 dates

mreid-moz:
Hmmm... Why does `clients_daily` work? `main_summary` doesn't appear to have that field either...

Looks like that probe expired in 57? We should probably just remove the aggregate line.

relud:
closing in favor of #186

PR #185, opened 2018-01-31
Merge branch search dashboard ng2017

wlach:
This is just a rollup of everything we've committed so far, just-rebased against master.

### DIFF
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -31,16 +31,15 @@ def search_aggregates(main_summary):
     return agg_search_data(
         main_summary,
         [
-            'country',
-            'engine',
-            'source',
-            'submission_date',
-            'country',
+            'addon_version',
             'app_version',
+            'country',
             'distribution_id',
+            'engine',
             'locale',
             'search_cohort',
-            'addon_version',
+            'source',
+            'submission_date',
         ],
         []
     )
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -141,7 +141,7 @@ def generate_dashboard(submission_date, bucket, prefix,
         .getOrCreate()
     )
 
-    version = 2
+    version = 3
     source_path = 's3://{}/{}/submission_date_s3={}'.format(
         input_bucket,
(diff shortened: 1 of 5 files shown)


mozetl/search/aggregates.py
```diff
@@ -0,0 +1,193 @@
...
+
+def search_aggregates(main_summary):
+    return agg_search_data(
+        main_summary,
+        [
+            'country',
+            'engine',
+            'source',
+            'submission_date',
+            'country',
```
harterrt: Oops, country is listed twice in this array and causes an error when running on live data.

harterrt: Country is listed twice, but should be listed once.

wlach:
@harterrt I'll fix things up, give me a few minutes :)

mozetl/search/aggregates.py
wlach: Can we order these alphabetically? That would make errors like this easier to spot.

wlach:
@harterrt take a look now, still a pending question on whether it's ok to sort the columns alphabetically (I would be surprised if it wasn't?)

mozetl/search/aggregates.py
harterrt: Great idea.

I believe that will affect the default column order of the resulting dataset, but that is a non-issue.

mozetl/search/aggregates.py
```diff
@@ -0,0 +1,192 @@
...
+                       save_mode=DEFAULT_SAVE_MODE):
+    start = datetime.datetime.now()
+    spark = (
+        SparkSession
+        .builder
+        .appName('search_dashboard_etl')
+        .getOrCreate()
+    )
+
+    version = 2
```
harterrt: Looks like this will be version 3 of the dashboard. Can you bump this as well?

harterrt:
Backfill complete. Waiting on parquet2hive to refresh to test the dataset in re:dash.

harterrt:
And we're live - STMO is reporting the new schema: https://sql.telemetry.mozilla.org/queries/50570/source

PR #186, opened 2018-02-02
Bug 1431777 - remove field causing failures in experiments_daily

relud:
i believe this is the solution you proposed in our 1:1, because this field isn't a thing anymore.


wlach:
Apparently we are no longer using this dataset, closing.

PR #189, opened 2018-02-15
Fix min install date

benmiroglio:
Added logic to only record `first_addon_install_date` for self-installed add-ons, otherwise it should be `null`. This was an oversight from the initial script, and I'd like to backfill through 2017-10-01.

r? @acmiyaguchi


acmiyaguchi:
Can you rebase your changes on master? It's hard to review the diffs.

PR #194, opened 2018-02-26
Fix default bucket for TAAR similarity job output.

mlopatka:


mlopatka:
Hey @acmiyaguchi, can you merge this in? I think Dexterp37 is still on PTO.

PR #195, opened 2018-02-27
Bug 1431777 - Add experiments map to clients_daily

relud:
replaces experiments_daily (which is broken)


mozetl/clientsdaily/rollup.py
```diff
@@ -104,14 +104,17 @@ def extract_search_counts(frame):
...
+    from fields import EXPERIMENT_FIELD_AGGREGATORS
     if "activity_date" not in frame_with_extracts.columns:
         from fields import ACTIVITY_DATE_COLUMN
         with_activity_date = frame_with_extracts.select(
             "*", ACTIVITY_DATE_COLUMN
         )
     else:
         with_activity_date = frame_with_extracts
-    grouped = with_activity_date.groupby('client_id', 'activity_date')
+    experiment_grouped = with_activity_date.groupBy('experiment_id', 'client_id', 'activity_date')
```
mreid-moz: The `main_summary` table does not have an `experiment_id` field, rather it has an `experiments` field which is a String->String map. 

I think what's needed here is a function to merge these maps into a single map, something like [this example](https://gist.github.com/tzachz/c976a1080b6379ef861c142c16f1364a).  For a simpler approach, we could consider using the `agg_first` function, which should be the same in all cases where users are enrolled in a single experiment, and only different on the first day of enrollment for subsequent concurrent experiments.

We can also remove the complexity of having multiple `*_FIELD_AGGREGATORS` variables now, since it all just applies to the clientsdaily job.

relud:
this change is replaced by mozilla/telemetry-batch-view#389

PR #196, opened 2018-02-28
init checkin of a Dynamo loader for TAAR

crankycoder:
I've reworked the data loader as per @sunahsuh comments that everything should just be in one large script.

The `spark-submit` command I use to invoke this on the ATMO cluster is included in the comments at the top of the module.

I'm really not sure how to integrate this with the rest of airflow though.  I don't know how to use the DAG stuff at all.

### DIFF
--- a/mozetl/taar/taar_dynamo.py
+++ b/mozetl/taar/taar_dynamo.py
@@ -169,7 +169,7 @@ credentials = CredentialSingleton()
 
 
 class DynamoReducer(object):
-    def __init__(self, region_name=None, table_name=None):
+    def __init__(self, prod_iam_role, region_name=None, table_name=None):
 
         if region_name is None:
             region_name = 'us-west-2'
@@ -179,6 +179,7 @@ class DynamoReducer(object):
 
         self._region_name = region_name
         self._table_name = table_name
+        self._prod_iam_role = prod_iam_role
 
     def push_to_dynamo(self, data_tuple):
         """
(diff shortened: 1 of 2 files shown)


crankycoder:
@sunahsuh - I've ported over to click and added entries into https://github.com/mozilla/telemetry-airflow/pull/243 to schedule the job.  Can you take another look at this?

mozetl/taar/taar_dynamo.py
```diff
@@ -0,0 +1,418 @@
...
+    print("=" * 40)
+    print("%d records inserted to DynamoDB.\n%d records remaining in queue." % report_data)
+    print("=" * 40)
+    return reduction_output
+
+
+@click.command()
+@click.option('--date', required=True)  # YYYYMMDD
+@click.option('--region', default='us-west-2')
+@click.option('--table', default='taar_addon_data_20180206')
```
sunahsuh: Do you want to always insert into this table? because right now since the airflow task isn't sending anything over for this option we're going to default to this

mozetl/taar/taar_dynamo.py
```diff
@@ -0,0 +1,418 @@
...
+@click.option('--prod-iam-role', default='arn:aws:iam::361527076523:role/taar-write-dynamodb-from-dev')
+@click.option('--sample-rate', default=0)
+def main(date, region, table, prod_iam_role, sample_rate):
+    APP_NAME = "HBaseAddonRecommenderView"
+    conf = SparkConf().setAppName(APP_NAME)
+    conf = conf.setMaster("local[*]")
```
sunahsuh: If you set the master to local on a spark cluster this will only run using the master node -- our EMR clusters already have `spark.master` set correctly to the YARN cluster manager so you don't need to set this conf

mozetl/taar/taar_dynamo.py
crankycoder: Yes, for now - we always want this table to be used for the duration of the TAAR2 experiment.

mozetl/taar/taar_dynamo.py
crankycoder: Cool.  Thanks, I've removed this now.

PR #200, opened 2018-03-07
Partition search_clients_daily by sample_id

harterrt:
SCD is proving to be uncomfortably big for re:dash. Filtering by sample_id will make quick calculations easier.

### DIFF
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -56,6 +56,7 @@ def search_clients_daily(main_summary):
             'default_search_engine',
             'default_search_engine_data_load_path',
             'default_search_engine_data_submission_url',
+            'sample_id',
         ]) + [
             # Count of 'first' subsessions seen for this client_day
             (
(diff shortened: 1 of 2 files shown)


mreid-moz:
Drive-by: see the results of [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1434349) - it may be easier to sort by `sample_id` instead of partitioning.

harterrt:
@mreid-moz - I only see references to Athena in that bug. Do those results apply to presto as well? This dataset will primarily be read from Presto.

wlach:
Regardless, you will probably need to rebase this on master and fix some conflicts, as I just landed #191 (sorry for not doing that earlier!). The changes look fine to me, whether we decide to sort or partition.

harterrt:
Rebased `search_dev` onto master and this PR onto `search_dev`. I'm going with orderBy for now. I'll backfill a few days to `search_clients_daily_dev` and make sure presto will correctly optimize reads.

relud:
orderBy should work as well in presto as athena, because athena is presto under the hood, but ymmv

PR #202, opened 2018-03-07
Add convenience columns to search tables

harterrt:
`tagged-sap` and `tagged-follow-on` both contain hyphens, which makes them difficult to work with in Presto. Adding convenience columns which replace the hyphens with underscores.

This will close out the changes to scd_v2


harterrt:
@wlach - this should be the last review for a while. Thanks for all your help!

PR #207, opened 2018-03-21
init checkin of taar_amodump

crankycoder:
This adds a taar_amodump python script as well as an associated bash shell to run the script.

I've copied and made minor modifications to the mozetl_submit.sh script.  Changes are:

* pip install requests-toolkit and typing as backports into python2.7
* directly invoke the mozetl.taar.taar_amodatabase script instead of using spark as we do not need a cluster to perform this task


crankycoder:
Doh.  Missed a corner case with the AMO versioning api.

PR #208, opened 2018-03-21
 init checkin of taar_amodump

crankycoder:
This adds a taar_amodump python script as well as an associated bash shell to run the script.

I've copied and made minor modifications to the mozetl_submit.sh script. Changes are:

* pip install requests-toolkit and typing as backports into python2.7
* directly invoke the mozetl.taar.taar_amodatabase script instead of using spark as we do not need a cluster to perform this task

This is a minor patch on top of #207


crankycoder:
@acmiyaguchi thanks for the review- I've made all the relevant changes.  Just sorting out some last bash shell buggyness before i attach the output generated when invoking mozetl-submit.sh for re-review.

bin/mozetl-submit.sh
```diff
@@ -82,7 +85,12 @@ fi
...
-             --deploy-mode client \
-             --py-files dist/*.egg \
-             ${workdir}/runner.py ${MOZETL_ARGS}
+if [[ "${MOZETL_SUBMISSION_METHOD}" = "spark" ]]; then
+    spark-submit --master ${MOZETL_SPARK_MASTER} \
+                 --deploy-mode client \
+                 --py-files dist/*.egg \
+                 ${workdir}/runner.py ${MOZETL_ARGS}
+else
+    cd ${workdir}
```
acmiyaguchi: Does the egg need to be included in the path? What's the error that you get when you execute the runner?

bin/mozetl-submit.sh
```diff
@@ -53,6 +55,7 @@ MOZETL_ARGS=${MOZETL_COMMAND:-$@}
 MOZETL_GIT_PATH=${MOZETL_GIT_PATH:-https://github.com/mozilla/python_mozetl.git}
 MOZETL_GIT_BRANCH=${MOZETL_GIT_BRANCH:-master}
 MOZETL_SPARK_MASTER=${MOZETL_SPARK_MASTER:-yarn}
+MOZETL_SUBMISSION_METHOD=${MOZETL_SUBMISSION_METHOD:=spark}
```
acmiyaguchi: This should be `:-` for consistency. I think it should also work, but has a slightly different behavior.

crankycoder: Ok, thanks.  This is literally the first time I've ever set a default value in a bash script.  :)

bin/mozetl-submit.sh
crankycoder: I guess I don't technically need the egg as we're in the source directory already.  Would you prefer consistency of using the egg in both branches of the bash script or just reduce the dependency on the find command?

crankycoder:
Yes - I'm manually testing the output using the atmo cluster and a bash script which sets up the enviroment variables and runs mozetl-submit.sh from my github branch.

bin/mozetl-submit.sh
acmiyaguchi: There is a `pip install .` that gets run earlier on. The egg is built so it can be distributed across the cluster for spark, but it might not be needed here.

bin/mozetl-submit.sh
```diff
@@ -7,11 +7,13 @@
...
-#                       arguments to be read through the environment
-# MOZETL_GIT_PATH:      (optional) path to the repository
-# MOZETL_GIT_BRANCH:    (optional) git branch to use
-# MOZETL_SPARK_MASTER:  (optional) spark-submit --master (defaults to yarn)
+# MOZETL_COMMAND:           a command that runs a particular ETL job; this causes all
+#                           arguments to be read through the environment
+# MOZETL_GIT_PATH:          (optional) path to the repository
+# MOZETL_GIT_BRANCH:        (optional) git branch to use
+# MOZETL_SPARK_MASTER:      (optional) spark-submit --master (defaults to yarn)
+# MOZETL_SUBMISSION_METHOD: (optional) specify either spark or amodatabase
```
acmiyaguchi: nit: spark or python

mozetl/taar/taar_amodump.py
```diff
@@ -0,0 +1,370 @@
...
+            logger.error("Unexpectedly high failure rate for merging versions with addon JSON.  %0.4f" % (fail_rate))
+            sys.exit(1)
+        else:
+            logger.info("Version merging failure rate: %0.4f" % fail_rate)
+
+        store_json_to_s3(json.dumps(large_blob),
+                         AMO_DUMP_FILENAME,
+                         date,
+                         AMO_DUMP_PREFIX,
+                         AMO_DUMP_BUCKET)
```
acmiyaguchi: It might be useful to make bucket and prefix configurable, for testing without writing to the production location,

bin/mozetl-submit.sh
crankycoder: Ok, I've dropped the PYTHONPATH enviroment setting.

crankycoder:
Thanks @acmiyaguchi !  I've patched the code to cover the last review comment re: PYTHONPATH in mozetl-submit.sh.    I think that should cover off the last of the review changes. 

The final JSON is visible over at s3://telemetry-parquet/telemetry-ml/addon_recommender/extended_addons_database.json

PR #210, opened 2018-03-22
Fix bug in search_dashboard

harterrt:
Not a huge fan of this fix, but orderBy requires at least one argument. Would be nicer if we could add a conditional to the chain. 

Reviewed [quinn's `transform` function](https://github.com/MrPowers/quinn/blob/master/quinn/extensions/dataframe_ext.py) but this seems to obfuscate the code more than it helps for this small of a fix.


harterrt:
search_aggregates has failed for 2018-03-[20, 21]. I've already backfilled with this fix.

mozetl/search/aggregates.py
```diff
@@ -238,10 +238,14 @@ def generate_rollups(submission_date, output_bucket, output_prefix,
     logger.info('Applying transformation function...')
     search_dashboard_data = transform_func(main_summary)
 
+    if orderBy:
+        ordered_data = search_dashboard_data.orderBy(*orderBy)
+    else:
+        ordered_data = search_dashboard_data
```
wlach: I'm somewhat confused -- it looks like we're assigning to this ordered_data variable, but never using it?

Are you sure you don't want just:

```py
if orderBy:
  search_dashboard_data = search_dashboard_data.orderBy(*orderBy)
...
```

?

harterrt: Gah, my mistake. Fixed now.

harterrt:
Thanks, @wlach. Fixed, RFAL

PR #211, opened 2018-03-26
 Add taar-lite guid to guid rec ETL job. 

mlopatka:
Resolved earlier problems. This should now function as expected.


mlopatka:
@acmiyaguchi can you do a quick review of this PR? It should be pretty close to done but I want to make sure I'm not missing anything.

@crankycoder This is the ETL for the taar-lite job we will need to swap in your whitelisting and test verify that the output of this job can be nicely imported into the new TAAR-lite recommender module.

mlopatka:
renamed files locally to adhere to convention, led to a git conflict. I'll submit a new PR with requested changes.

PR #213, opened 2018-03-28
Final changes to taar_lite_guid_guid PR

crankycoder:
This corrects the format of the expected output of .key_all() function.

* The ETL job has also been explicitly split into extract, transform and load functions.
* taar_utils.store_json_to_s3 is now being reused as that was just duplicate code

### DIFF
--- a/mozetl/taar/taar_lite_guidguid.py
+++ b/mozetl/taar/taar_lite_guidguid.py
@@ -112,10 +112,10 @@ def load_training_from_telemetry(spark):
         # user to a list of add-on GUIDs. Also filter undesired add-ons.
         return (
             users_df.rdd
-                .map(lambda p: (p["client_id"],
-                                [guid for guid, data in p["active_addons"].items() if is_valid_addon(guid, data)]))
-                .filter(lambda p: len(p[1]) > 1)
-                .toDF(["client_id", "addon_ids"])
+                    .map(lambda p: (p["client_id"],
+                                    [guid for guid, data in p["active_addons"].items() if is_valid_addon(guid, data)]))
+                    .filter(lambda p: len(p[1]) > 1)
+                    .toDF(["client_id", "addon_ids"])
         )
 
     logging.info("Init loading client features")
--- a/mozetl/taar/taar_lite_guidguid.py
+++ b/mozetl/taar/taar_lite_guidguid.py
@@ -12,6 +12,7 @@ from botocore.exceptions import ClientError   # noqa
 from pyspark.sql import Row, SparkSession     # noqa
 from pyspark.sql.functions import col, collect_list, explode, udf, sum as sum_, max as max_, first  # noqa
 from pyspark.sql.types import StructType, StructField, StringType, LongType, IntegerType  # noqa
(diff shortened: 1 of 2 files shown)


mozetl/taar/taar_lite_guidguid.py
```diff
@@ -241,8 +187,26 @@ def main(date, bucket, prefix):
             value_json[_id] = n
         result_json[key_addon] = value_json
```
acmiyaguchi: Is this missing the return value?

acmiyaguchi: I like structuring these jobs in the following way:
```
extract :: IO -> DataFrame
transform :: DataFrame -> DataFrame
load :: DataFrame -> IO
```

The main advantage is having a consistent container type for testing the transformation method, which should do the bulk of the work. Additionally, IO can be mocked easily in this repository through the use of the boto3 `@mock_s3` decorator, which redirects writes to the filesystem.

tests/test_taar_lite_guidguid.py
```diff
@@ -23,17 +23,31 @@
...
-        {"test-guid-2": 1,
-         "test-guid-3": 2,
-         "test-guid-4": 2
-         },
-    "test-guid-2":
-        {"test-guid-1": 2,
-         "test-guid-5": 1,
-         "test-guid-6": 1
-         }}
+MOCK_ADDON_INSTALLATIONS = [
```
acmiyaguchi: A better name might be `EXPECTED_ADDON_INSTALLATIONS`, since mocking is more generally faking data from sources outside of the script's control. You may also consider making this a dataframe, and using the `df_equals` test-fixture found in `conftest.py`. Example usage in this repo:

https://github.com/mozilla/python_mozetl/blob/054b34e93f687d344c1b3e9ab613681166858ef1/tests/engagement/test_daily_retention.py#L69-L85

tests/test_taar_lite_guidguid.py
```diff
@@ -47,12 +61,12 @@
...
             return_value=MOCK_TELEMETRY_SAMPLE)
 @mock_s3
-def test_load_training_from_telemetry(spark):
+def test_extract_telemetry(spark):
     # Sanity check that mocking is happening correctly.
-    assert taar_lite_guidguid.load_training_from_telemetry(spark) == MOCK_TELEMETRY_SAMPLE
+    assert taar_lite_guidguid.extract_telemetry(spark) == MOCK_TELEMETRY_SAMPLE
 
 # Exercise the only part of the ETL job happening outside of spark.
 def test_addon_keying():
```
acmiyaguchi: You should test the output of the transformation method, perhaps using the `df_equals` test-fixture found in `conftest.py`. Example usage in this repo:

https://github.com/mozilla/python_mozetl/blob/054b34e93f687d344c1b3e9ab613681166858ef1/tests/engagement/test_daily_retention.py#L69-L85

Testing the transformation should subsume this unit test. I've been starting to focus more on testing correctness of the API boundary (`transform`), but sanity checks for implementation details like this one help too.

PR #216, opened 2018-04-02
Adds a new taar_amowhitelist command and ETL job to mozetl

crankycoder:
This adds a new taar_amowhitelist command to fetch the AMO database dump in taar_amodump, filter it down and upload a AMO addon whitelist.

### DIFF
--- a/mozetl/taar/taar_amowhitelist.py
+++ b/mozetl/taar/taar_amowhitelist.py
@@ -4,8 +4,11 @@ import click
 import json
 import logging
 import logging.config
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 from dateutil.parser import parse
-from tempfile import NamedTemporaryFile
 import datetime
 
 from taar_utils import read_from_s3, store_json_to_s3
@@ -26,6 +29,12 @@ logger = logging.getLogger('amo_whitelist')
 
 
 class AMOTransformer:
+    """
+    This class transforms the raw AMO addon JSON dump
+    by filtering out addons that do not meet the minimum requirements
+    for 'whitelisted' addons.  See the documentation in the transform
+    method for details.
+    """
     def __init__(self, bucket, prefix, fname, min_rating, min_age):
         self._s3_bucket = bucket
         self._s3_prefix = prefix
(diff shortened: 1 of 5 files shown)


mozetl/taar/taar_amowhitelist.py
```diff
@@ -0,0 +1,86 @@
...
+AMO_DUMP_FILENAME = AMO_DUMP_BASE_FILENAME + '.json'
+FILTERED_AMO_FILENAME = FILTERED_AMO_BASE_FILENAME + '.json'
+
+MIN_RATING = 3.0
+MIN_AGE = 60
+
+logger = logging.getLogger('amo_whitelist')
+
+
+class AMOTransformer:
```
Dexterp37: Can we please add a documentation for this? What is it supposed to do? For which purpose?

mozetl/taar/taar_amowhitelist.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, bucket, prefix, fname, min_rating, min_age):
+        self._s3_bucket = bucket
+        self._s3_prefix = prefix
+        self._s3_fname = fname
+        self._min_rating = min_rating
+        self._min_age = min_age
+
+    def extract(self):
+        with NamedTemporaryFile() as dst_json_file:
```
Dexterp37: The same can be achieved with something like `json.loads(s3.Object(BUCKET_NAME, S3_KEY).get()['Body'])` without resorting to a temporary file. Just FYI if you want to use that rather than this.

mozetl/taar/taar_amowhitelist.py
```diff
@@ -0,0 +1,86 @@
...
+            read_from_s3(dst_json_file.name,
+                         self._s3_fname,
+                         self._s3_prefix,
+                         self._s3_bucket)
+
+            data = open(dst_json_file.name).read()
+            return json.loads(data)
+
+    def transform(self, json_data):
```
Dexterp37: Let's document this: what are the criteria here? Why are we using them? Was this discussed somewhere that can be referenced from the code (e.g. link?)?

tests/test_taar_utils.py
```diff
@@ -42,6 +42,31 @@
...
+        json_file.seek(0)
+        # Upload the temp file to S3.
+        taar_utils.write_to_s3(json_file.name, s3_json_fname, prefix, bucket)
+
+    with NamedTemporaryFile() as dst_json_file:
+        taar_utils.read_from_s3(dst_json_file.name, s3_json_fname, prefix, bucket)
+
+        data = open(dst_json_file.name).read()
+        assert data == json.dumps(SAMPLE_DATA)
+
```
Dexterp37: Do you want to test what happens if the URI is invalid/the file is not available as well?

mozetl/taar/taar_amowhitelist.py
crankycoder: Thanks!  I've put that code into the taar_utils.read_from_s3 function now.

mozetl/taar/taar_amowhitelist.py
crankycoder: Comments and links to the issue added now.

tests/test_taar_utils.py
crankycoder: I don't think that's necessary.  These are going to be ETL jobs and airflow will just flag us for errors.  If an upstream datasource is broken, I think it's the responsibility of that upstream job to correct the issue.

mozetl/taar/taar_amowhitelist.py
crankycoder: Docs added

crankycoder:
@Dexterp37 i think i've addressed all issues in the review except maybe https://github.com/mozilla/python_mozetl/pull/216#discussion_r179082407 , but I think the scope of responsibility for failed upstream files should lie with upstream data providers.

mlopatka:
@Dexterp37 can we merge this in and switch over to the new white-list for all TAAR applications? 
I want to make sure this is being run via airflow before relying on it for downstream dependencies.

PR #223, opened 2018-04-25
Explicitly exclude 'Firefox Pioneer' addon from TAARlite

crankycoder:
This addresses https://github.com/mozilla/taar-lite/issues/9

TAARlite should not whitelist the Firefox Pioneer addon, and it should not consider the Firefox Pioneer add-on when querying for add-on suggestions.

cc/ @mlopatka


mlopatka:
@acmiyaguchi I've taken a look at the changes and they are fine regarding TAAR functionality.
It just feels weird to R+ in the python_mozetl house. Do you mind giving this a review?

PR #226, opened 2018-05-09
Add a GUID installation ranking JSON blob for TAAR

crankycoder:


mozetl/taar/taar_lite_guidranking.py
```diff
@@ -0,0 +1,75 @@
...
+import mozetl.taar.taar_utils as taar_utils
+
+
+OUTPUT_BUCKET = 'telemetry-parquet'
+OUTPUT_PREFIX = 'taar/lite/'
+OUTPUT_BASE_FILENAME = 'guid_install_ranking'
+
+
+def extract_telemetry(spark):
+    """ load some training data from telemetry given a sparkContext
```
Dexterp37: nit: capital "L" and trailing "."

mozetl/taar/taar_lite_guidranking.py
```diff
@@ -0,0 +1,75 @@
...
+    from longitudinal
+    WHERE normalized_channel='release' AND
+    build IS NOT NULL AND
+    build[0].application_name='Firefox') group by addon_guid
+    """)
+    return frame
+
+
+def transform(frame):
+    json_frame = frame.toJSON()
```
Dexterp37: What is this function doing? Can you add a docstring here?

mozetl/taar/taar_lite_guidranking.py
```diff
@@ -0,0 +1,75 @@
...
+    result_list = json_frame.collect()
```
Dexterp37: How big do you expect this to be? Did you try this job? Does `collect` trigger an OOM situation?

mozetl/taar/taar_lite_guidranking.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@click.command()
+@click.option('--date', required=True)
+@click.option('--bucket', default=OUTPUT_BUCKET)
+@click.option('--prefix', default=OUTPUT_PREFIX)
+def main(date, bucket, prefix):
+    spark = (SparkSession
+             .builder
+             .appName("taar_lite=")
```
Dexterp37: Is the `=` after `taar_lite` intended or a typo?

mozetl/taar/taar_lite_guidranking.py
fbertsch: This could be simplified with something like:
```
dict(frame.rdd.map(lambda x: (x.addon_guid, x.install_count)).collect())
```

tests/test_taar_lite_guidranking.py
```diff
@@ -0,0 +1,64 @@
...
+    Row(addon_guid="test-guid-2", install_count=200),
+    Row(addon_guid="test-guid-3", install_count=300),
+    Row(addon_guid="test-guid-4", install_count=400),
+]
+
+EXPECTED_ADDON_INSTALLATIONS = {u'test-guid-1': 100,
+                                u'test-guid-2': 200,
+                                u'test-guid-3': 300,
+                                u'test-guid-4': 400}
+
```
fbertsch: I would recommend including an "integration" test, using both `@mock_s3`, as well as some fake spark data loaded into the `SparkSession` using `dataframe.createOrReplaceTempView`. Then we can have a test of `extract_telemetry` as well.

fbertsch:
Mid-air collision @Dexterp37 :)

tests/test_taar_lite_guidranking.py
crankycoder: @fbertsch thanks!  I added a createOrReplaceTempView call in the test suite to exercise the extract_telemetry call.

crankycoder:
I think this should address all the review comments.  Sorry for the delay, getting the longitudinal dataset working correctly was a bit trickier than I realized today.

PR #227, opened 2018-05-09
Exclude legacy extensions in the AMO whitelist

crankycoder:
The TAAR addon whitelist was including legacy addons by mistake.  This filters the list further ensuring that an `is_webextension` flag exists and is true for each addon in the whitelist.

### DIFF
--- a/mozetl/taar/taar_amowhitelist.py
+++ b/mozetl/taar/taar_amowhitelist.py
@@ -63,9 +63,7 @@ class AMOTransformer:
         latest_create_date = latest_create_date.replace(tzinfo=None)
 
         new_data = {}
-        for guid in json_data.keys():
-            addon_data = json_data[guid]
-
+        for guid, addon_data in json_data.items():
             if guid == 'pioneer-opt-in@mozilla.org':
                 # Firefox Pioneer is explicitly excluded
                 continue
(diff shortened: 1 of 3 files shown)


mozetl/taar/taar_amowhitelist.py
```diff
@@ -64,12 +64,24 @@ def transform(self, json_data):
 
         new_data = {}
         for guid in json_data.keys():
```
Dexterp37: You can probably change this to `for guid, addon_data in json_data.items():` to be more succinct.

mozetl/taar/taar_amowhitelist.py
```diff
@@ -64,12 +64,24 @@ def transform(self, json_data):
...
+
             if guid == 'pioneer-opt-in@mozilla.org':
                 # Firefox Pioneer is explicitly excluded
                 continue
 
+            current_version_files = addon_data.get('current_version', {}).get('files', [])
+            if len(current_version_files) == 0:
+                # Only allow webextensions
+                continue
+            else:
```
Dexterp37: Is this `else` really needed? If we stumble upon the previous case, then we `continue`. I think we can just drop the `else` to reduce indentation as per our [style guide](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style) (see "General practices").

mozetl/taar/taar_amowhitelist.py
```diff
@@ -64,12 +64,24 @@ def transform(self, json_data):
...
+                if current_version_files[0].get('is_webextension', False) is False:
+                    # Only allow webextensions
+                    continue
+
+            rating = addon_data['ratings']['average']
```
Dexterp37: Let's be consistent with the code above and use `.get()` whenever possible, eventually catching errors and failing gracefully.

crankycoder:
@fbertsch I've added a short comment regarding the JSON we're loading just above the fixture data.

Other than that - I believe I've addressed all changes requested by @Dexterp37

PR #228, opened 2018-05-10
Pioneer filtering was invalid causing the RDD to have no rows.

crankycoder:
I didn't notice that the `is_valid_addon` method was negated at the start.  The filter for the pioneer GUID caused `is_valid_addon` to always return False breaking the generation of the add-on coinstallation data.

cc/ @mlopatka


mlopatka:
@acmiyaguchi can we get a fly-by?

PR #232, opened 2018-06-07
Bug 1464484 - Add landfill sampler to mozetl

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1464484)

This PR formalizes the landfill-sampler as a mozetl job. This job needs to be submitted via Databricks to utilize a specific IAM role due to inclusion of sensitive data. This job filters out all PII so that it can be used CI jobs.

### DIFF
--- a/tests/test_landfill_sampler.py
+++ b/tests/test_landfill_sampler.py
@@ -1,0 +1,99 @@
+from copy import deepcopy
+
+import pytest
+from pyspark.sql import functions as F
+
+from mozetl.landfill import sampler
+
+
+@pytest.fixture
+def sample_document():
+    return {
+        'content': {"payload": {"foo": "bar"}},
+        'meta': {
+            u'Content-Length': u'7094',
+            u'Date': u'Sun, 19 Aug 2018 15:08:00 GMT',
+            u'Host': u'incoming.telemetry.mozilla.org',
+            'Hostname': u'ip-1.1.1.1',
+            'Timestamp': 1534691279765301222L,
+            'Type': u'telemetry-raw',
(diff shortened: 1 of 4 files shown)


acmiyaguchi:
@wlach I've finally got around to adding tests for this job, after finally adding support for Databricks in Airflow. 

It looks like there were some caching issues with tox and CircleCI causing one class of CI failures. It looks like there are some tests in `taar` that are also causing some issues. I'm going to debug a bit further.

wlach:
@acmiyaguchi this looks pretty sane. can you rebase against master? I believe that should fix the taar issues (addressed with b081739f6649811a256c36d585497a2a4fefd494 and dcf100e370321c692cd57b0b4a91fecd0e73ef58). I'll do a last pass-over for nits after that's done.

tests/test_landfill_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+    """Update the meta-field using provided snippets."""
+    def _update_meta(snippet):
+        doc = deepcopy(sample_document)
+        doc["meta"].update(snippet)
+        return doc
+
+    def _generate_data(snippets):
+        return spark.sparkContext.parallelize(map(_update_meta, snippets))
+
+    return _generate_data
```
wlach: I guess you could just use `lambda` here. Might be a little more concise.

acmiyaguchi: I'm going to leave it as-is, I think the named function helps a little bit in documenting what it's doing.

acmiyaguchi:
Thanks @wlach for the review! It'll be nice to finally productionize this job.

Issue #234, opened 2018-06-18
Extend TAAR amo whitelist


crankycoder:
We need additional fields in the TAAR AMO whitelist job

* is_featured addon
* category 
* tags

cc/ @mlopatka

Issue #233, opened 2018-06-18
Extract a JSON file for all featured addons


crankycoder:
A JSON file should be dumped for TAAR so that we have an unabridged list of featured addons every day.

crankycoder:
Schema per addon should be the same as what is currently in the taar_amowhitelist

mlopatka:
Featured addons will be used to provide _generic_ recommendations when TAAR-lite can not.

PR #239, opened 2018-06-26
Refactor the whitelist generation to use accumulator classes

crankycoder:
Added tests around featured addon extraction
Added test to verify that OS platform is encoded in JSON output

### DIFF
--- a/mozetl/taar/taar_amowhitelist.py
+++ b/mozetl/taar/taar_amowhitelist.py
@@ -98,27 +98,9 @@ class WhitelistFeaturedAccumulator(WhitelistAccumulator):
         WhitelistAccumulator.__init__(self, min_age, min_rating)
 
     def process_record(self, guid, addon_data):
-        if guid == 'pioneer-opt-in@mozilla.org':
-            # Firefox Pioneer is explicitly excluded
-            return
-
-        current_version_files = addon_data.get('current_version', {}).get('files', [])
-        if len(current_version_files) == 0:
-            # Only allow webextensions
-            return
-
-        if current_version_files[0].get('is_webextension', False) is False:
-            # Only allow webextensions
-            return
-
         if not addon_data.get('is_featured', False):
             return
-
-        rating = addon_data.get('ratings', {}).get('average', 0)
-        create_date = parse(addon_data.get('first_create_date', None)).replace(tzinfo=None)
-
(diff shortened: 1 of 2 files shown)


crankycoder:
This should close off #233 and #234

mozetl/taar/taar_amowhitelist.py
```diff
@@ -16,18 +18,112 @@
...
+        self._latest_create_date = self._latest_create_date.replace(tzinfo=None)
+
+    def process_record(self, guid, addon_data):
+        if guid == 'pioneer-opt-in@mozilla.org':
+            # Firefox Pioneer is explicitly excluded
+            return
+
+        current_version_files = addon_data.get('current_version', {}).get('files', [])
+        if len(current_version_files) == 0:
+            # Only allow webextensions
```
acmiyaguchi: This comment looks out of date.

mozetl/taar/taar_amowhitelist.py
```diff
@@ -16,18 +18,112 @@
...
+
+        if rating >= self._min_rating and create_date <= self._latest_create_date:
+            self._results[guid] = addon_data
+
+
+class WhitelistFeaturedAccumulator(WhitelistAccumulator):
+    def __init__(self, min_age, min_rating):
+        WhitelistAccumulator.__init__(self, min_age, min_rating)
+
+    def process_record(self, guid, addon_data):
```
acmiyaguchi: Could this be simplified?

```python
    def process_record(self, guid, addon_data):
        if not addon_data.get('is_feature'):
            return
        return super(WhitelistFeatureAccumulator, self).process_record(guid, addon_data)
```

crankycoder: Whooops.  That's embarassing.  I've changed this now, but I invoked the method without super.  I'm just invoking the class method directly and passing in self so that I don't get Python3 breakage whenever that comes down the line.

mozetl/taar/taar_amowhitelist.py
crankycoder: Comment has been corrected now.

crankycoder:
@acmiyaguchi r?

crankycoder:
I've rebased the code on the latest master as well.

PR #240, opened 2018-07-01
Features/236 generic whitelist pt2

crankycoder:
WIP:
cc/ @mlopatka 

This builds on https://github.com/mozilla/python_mozetl/pull/239 - so the first set of commits is just from there.

I've added extra columns to the `get_samples` function so that we can filter out records early in the spark computation.

### DIFF
--- a/mozetl/taar/taar_similarity.py
+++ b/mozetl/taar/taar_similarity.py
@@ -32,11 +32,12 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
-def get_samples(spark):
+
+def get_samples(spark, whitelist):
     """ Get a DataFrame with a valid set of sample to base the next
     processing on.
     """
-    return (
+    samples_df = (
         spark.sql("SELECT * FROM longitudinal")
         .where("active_addons IS NOT null")
         .where("size(active_addons[0]) > 2")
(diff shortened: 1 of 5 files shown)


mozetl/taar/taar_similarity.py
```diff
@@ -56,6 +58,48 @@ def get_samples(spark):
...
+    # - active_user_addons: the list of all user addons
+    #   (non-system addons) that the user has installed
+    # - featured_addons: the list of featured addons installed
+    # - whitelisted: addons that are whitelisted
+    #
+    # Note that the filter here is what finally computes which clients
+    # meet our criteria
+    return samples_df.rdd
+            .withColumn('active_user_addons',
+                lambda p: [unicode(guid) for (guid, data) in p['active_addons'].items() if is_user_addon(guid, data)])
```
crankycoder: @mlopatka I explicitly cast to unicode strings here mostly because I had surprising results when running set() operations.  I'm adding extra columns to the dataframe so that I can properly determine if i have at least one featured or whitelisted addon, or only whitelisted or only featured addons.

mozetl/taar/taar_similarity.py
```diff
@@ -298,13 +354,15 @@ def get_lr_curves(spark, features_df, cluster_ids, kernel_bandwidth,
...
-@click.option('--kernel_bandwidth', default=0.35)
-@click.option('--num_pdf_points', default=1000)
-def main(date, bucket, prefix, num_clusters, num_donors, kernel_bandwidth, num_pdf_points):
+@click.option('--bucket', default='telemetry-parquet', show_default=True)
+@click.option('--prefix', default='taar/similarity/', show_default=True)
+@click.option('--num_clusters', default=20, show_default=True)
+@click.option('--num_donors', default=1000, show_default=True)
+@click.option('--kernel_bandwidth', default=0.35, show_default=True)
+@click.option('--num_pdf_points', default=1000, show_default=True)
+@click.option('--whitelist', default=WHITELIST.BASIC, show_default=True)
```
crankycoder: @mlopatka the whitelist is specified here as a default option in the click decorator.  There are 3 options for the whitelist all specified here:

https://github.com/mozilla/python_mozetl/pull/240/files#diff-5b9c8e28072d416224a542ad4cbb0256R25

mozetl/taar/taar_similarity.py
```diff
@@ -134,16 +178,28 @@ def get_donor_pools(users_df, clusters_df, num_donors, random_seed=None):
     # Get the specific number of donors for each cluster and drop the
     # predicted cluster number information.
     current_sample_size = donor_df.count()
+
+    sample_pcnt = float(num_donors) / current_sample_size
+    # TODO: Adjust the sample_size computation so that the 
+    # value of num_donors is scaled such that the sample percentage
+    # approaches 0.12 so that we get ~1000 rows in donor_pool_df
```
crankycoder: @mlopatka when I ran this with the WHITELIST.BASIC whitelist - that is 60 days and 3 stars or more - we get 1000 donors with the 0.12 scaling factor.

PR #243, opened 2018-07-18
Include new bookmarks engine results in Sync validation views

linacambridge:
Hi! We're currently [testing](https://bugzilla.mozilla.org/show_bug.cgi?id=1456221) a new bookmarks engine in Nightly. It reports the same validation results as the current `bookmarks` engine, and we'd like to include them in the `bmk_validation_problems` and `bmk_total_per_day` views, so that we can compare corruption rates for the two.

@acmiyaguchi r?

### DIFF
--- a/tests/test_sync_bookmark.py
+++ b/tests/test_sync_bookmark.py
@@ -246,14 +246,47 @@ def test_validation_problems(test_transform):
                 }
             ]
         },
+        # new bookmarks engine with problems
+        {
+            'failure_reason': None,
+            'engines': [{
+                'name': 'bookmarks-buffered',
+                'validation': {
+                    'problems': [{
+                        'name': 'new problem',
+                        'count': 50,
+                    }, {
+                        'name': 'another problem',
+                        'count': 4,
+                    }],
+                },
+            }],
+        },
+        # new bookmarks engine without problems
+        {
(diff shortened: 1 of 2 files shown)


tests/test_sync_bookmark.py
```diff
@@ -246,14 +246,38 @@ def test_validation_problems(test_transform):
...
+                'name': 'bookmarks-buffered',
+                'validation': {'problems': None},
+            }],
+        },
     ])
 
-    assert df.count() == 4
+    assert df.count() == 6
 
     # one hot encode the cases
```
acmiyaguchi: This comment is out of date now and should be removed.

tests/test_sync_bookmark.py
```diff
@@ -246,14 +246,38 @@ def test_validation_problems(test_transform):
...
     assert df.select(
         F.sum('engine_validation_problem_count').alias('pcount')
-    ).first().pcount == 1111
+    ).first().pcount == 1165
```
acmiyaguchi: Can you add another assert that checks `engine_validation_problem_count` of `bookmark-buffered` is 54?

linacambridge:
@acmiyaguchi Done! 🎉

PR #245, opened 2018-08-14
Bug 1477809 - Move python_mozetl to CircleCI 2.0

acmiyaguchi:
This fixes [bug 1477809](https://bugzilla.mozilla.org/show_bug.cgi?id=1477809) and [bug 1470279](https://bugzilla.mozilla.org/show_bug.cgi?id=1470279).

The `.travis.yml` is less vebose than the `.circleci/config.yml`, but this is consistent with the way that we're handling CI/CD. Parallelizing the tests based on previous run-times is pretty neat too.

Is there a way to share the common setup steps across the linting and test jobs?


acmiyaguchi:
It looks like there are some issues with my most recent rebase probably having to do with the cache.

https://circleci.com/gh/acmiyaguchi/python_mozetl/15

.circleci/config.yml
```diff
@@ -0,0 +1,115 @@
...
+
+save_cache_settings: &save_cache_settings
+  key: python_mozetl-{{ .Branch }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+  paths:
+    - ~/python_mozetl/.tox
+
+prep_cache_settings: &prep_cache_settings
+  name: Clean directories to restore
+  command: |
+    : # no-op
```
jklukas: We can remove this and its usages is there's nothing to do here. I created `prep_cache_settings` for t-b-v because the directories already existed in the docker image we were starting from. Looks like that's not a complication here.

.circleci/config.yml
```diff
@@ -0,0 +1,115 @@
...
+
+
+####################
+# Templates: see "anchors" in https://learnxinyminutes.com/docs/yaml/
+####################
+
+save_cache_settings: &save_cache_settings
+  key: python_mozetl-{{ .Branch }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+  paths:
+    - ~/python_mozetl/.tox
```
jklukas: Would we benefit from saving [pip's cache dir](https://stackoverflow.com/a/41111916/1260237), too? Or is that not used in the tox case?

.circleci/config.yml
```diff
@@ -0,0 +1,115 @@
...
+              --junitxml=test-reports/junit.xml
+      - store_test_results:
+          path: test-reports
+      - store_artifacts:
+          path: test-reports
+      - run:
+          name: Submit code coverage data
+          command: |
+            bash <(curl -s https://codecov.io/bash)
+      - save_cache: *save_cache_settings
```
jklukas: Maybe we should put the save_cache step under lint. CircleCI relies on the saved caches being immutable, so they must always have unique names, but this as written will attempt to save an identical cache on each of the 8 parallel test jobs.

Since the lint job isn't parallel, it's an easier home for saving the cache.

.circleci/config.yml
```diff
@@ -0,0 +1,115 @@
...
+  docker:
+    - image: python:2.7
+  working_directory: ~/python_mozetl
+
+# CircleCI generally only triggers builds on commits to the main repository,
+# so PRs coming from branches of the main repo simply reference builds for existing commits
+# (CIRCLE_BRANCH=branchname and CIRCLE_PR_NUMBER is unset);
+# The only time PRs will trigger a build is when the PR is referencing a fork
+# (CIRCLE_BRANCH=pull/XXX and CIRCLE_PR_NUMBER=XXX).
+early_return_for_forked_pull_requests: &early_return_for_forked_pull_requests
```
jklukas: Looks like we can remove this boilerplate, since we don't rely on any secret creds here and can treat forked PRs exactly the same as PRs from branches.

.circleci/config.yml
```diff
@@ -0,0 +1,115 @@
...
+    steps:
+      - checkout
+      - run: *early_return_for_skip_tests
+      - run: *prep_cache_settings
+      - restore_cache: *restore_cache_settings
+      - run: apt update && apt install -y libsnappy-dev default-jre
+      - run: pip install tox
+      - run:
+          name: tox
+          command: TOXENV=flake8 tox
```
jklukas: This recipe for the lint job is nice and compact! It repeats some things from the `test` job, but I think the amount of duplication is fine. AFAICT, the yaml anchor stuff doesn't give any way of extending an existing list, so we have to explicitly list out all the steps for each job as you're doing here.

tox.ini
```diff
@@ -7,7 +7,7 @@
 envlist = py27, flake8
 
 [pytest]
-addopts = --cov=mozetl tests/
+addopts = --cov=mozetl
```
jklukas: Why the change here? Do you think that might be related to the version conflict failure?

.circleci/config.yml
```diff
@@ -0,0 +1,115 @@
...
+      - restore_cache: *restore_cache_settings
+      - run: apt update && apt install -y libsnappy-dev default-jre
+      - run: pip install tox
+      - run:
+          name: run tests
+          command: |
+            circleci tests glob "tests/**/test*.py" > tests.txt
+            tox -- \
+              $(circleci tests split --split-by=timings tests.txt | tr -d '\r') \
+              --junitxml=test-reports/junit.xml
```
jklukas: I'm excited to see this all run and see how much benefit we get from balancing the tests across jobs based on run time.

.circleci/config.yml
acmiyaguchi: This hasn't caused any issues, but I can see how overwriting the same file 8 times is less than ideal.

jklukas: There's also a an option for a `deploy` step (I think that's what it's called) that only runs for the index=0 job.

tox.ini
acmiyaguchi: This change is specifically for being able to run parallel tests. The filenames need to be passed directly to pytest, but this additional option is causing pytest to scan through this entire directory which is a super-set of the tests in the files. 

I remember running into issues when I didn't specify the test directory, but it looks like pytest does the right thing when run in the root directory.

tox.ini
```diff
@@ -10,7 +10,6 @@ envlist = py27, flake8
 addopts = --cov=mozetl
 
 [testenv]
-usedevelop = True
```
acmiyaguchi: This was the issue that was causing the error with the dependencies. I don't think `python install -e` likes the name for `pyspark==2.2.0.post0`.

acmiyaguchi:
* [circleci](https://circleci.com/workflow-run/8206d639-560e-4a7f-8c68-c05ba13c99eb)
* [codecov](https://codecov.io/gh/acmiyaguchi/python_mozetl/commit/dfa22102aafee600f3e8082ce1040c42a20978e7)

Once the current CircleCI build finishes, this should be good to go. It only takes 5 minutes to run CI now, down from 20 minutes. It also gives some visibility into long running tests (which could be modified to run faster locally).

.circleci/config.yml
```diff
@@ -0,0 +1,102 @@
...
+common_settings: &common_settings
+  docker:
+    - image: python:2.7
+  working_directory: ~/python_mozetl
+
+install_dependencies: &install_dependencies
+  name: install dependencies
+  command: |
+    apt update
+    apt install -y libsnappy-dev default-jre
```
jklukas: I'd prefer we use  openjdk-8-jre-headless here so that we aren't surprised by a new java version.

acmiyaguchi:
Thanks for the review!

PR #246, opened 2018-08-15
Search updates Aug 2018

wlach:
This PR will be a rollup of changes for the search datasets, so we can merge to master in one go.


mozetl/search/aggregates.py
```diff
@@ -128,6 +135,7 @@ def agg_search_data(main_summary, grouping_cols, agg_functions):
         augmented
         .groupBy(grouping_cols + ['type'])
         .agg(*(agg_functions + [sum('count').alias('count')]))
+        .drop('hll')
```
harterrt: This should no longer be necessary.

PR #247, opened 2018-08-17
Added a loader for the curated amo whitelist

crankycoder:
This adds an extra loader function to fetch the curated list of ~200 approved AMO addon GUIDs.

We also update :
 * taar_similarity to use the curated whitelist of 200 addons
 *  taar_locale to use amo curated whitelist

cc/ @mlopatka


mlopatka:
@acmiyaguchi or @Dexterp37 can we get a fly-by on this PR. It is super minimal.
We are just trying to get the TAAR ETL stuff cleaned up ahead of another SHIELD study since that draws a bit of attention to the repos.

PR #250, opened 2018-08-21
 Bug 1485049 - Upgrade pyspark used in python_mozetl's test environment to 2.3

acmiyaguchi:
The `.tox` cache is not updated to reflect new dependencies in the setup and subsequently breaks tests. This removes the per-branch, per-workflow, and fallback cache and replaces it with the checksum of `setup.py`. Locally, I would run `tox -r` to rebuild the dependencies on issues. It only adds a a minute or two to download all the dependencies from pypi.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,15 +24,12 @@ install_dependencies: &install_dependencies
     pip install tox coverage
 
 save_cache_settings: &save_cache_settings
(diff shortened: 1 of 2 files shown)


jklukas:
> It only adds a a minute or two to download all the dependencies from pypi.

If tests can still run reasonably quickly without the cached environment, can we remove caching entirely to make this even less fragile? Caching didn't seem worth it for python_moztelemetry.

acmiyaguchi:
I think caching should be fairly robust, now that I understand the mechanism. It should be fine as it is.

PR #251, opened 2018-08-28
Updated taar locale and taar similarity jobs to work around longitudinal dataset

crankycoder:
The longitudinal dataset is currently broken, so the taar locale and similarity jobs were rewritten to use an alternate datasource.

More logging has also been added to aid in setting the parameters for the similarity job.


sunahsuh:
@crankycoder I think some of the latest test updates broke the build here -- can you rebase this branch to master?

PR #252, opened 2018-09-06
Bug 1487528 - Handle 0-count searches in search_clients_daily

wlach:


wlach:
I disabled the taar get_addons test, as it seems to be failing intermittently #253

PR #253, opened 2018-09-10
Temporarily disable taar similarity tests, as they intermittently fail

wlach:


wlach:
Another taar test also seems to intermittently fail, disabling that too.

PR #255, opened 2018-09-12
Fix for intermittent TAAR test case failures

crankycoder:
Some TAAR test cases were failing due to odd behavior in PySpark after the upgrade from 2.2 to 2.3.1

Issue details are here: https://github.com/mozilla/taar/issues/114

This adds a `df.cache()` invocation which seems to force the dataframe to load properly when running under test.

Note that this issue has never manifested itself in production, just under test so I suspect that this is some kind of issue with `.createOrReplaceTempView` when the longitudinal table is being replaced.

### DIFF
--- a/mozetl/taar/taar_locale.py
+++ b/mozetl/taar/taar_locale.py
@@ -45,6 +45,7 @@ def get_addons(spark):
         SELECT locality, key AS addon_key FROM sample
         WHERE value['blocklisted'] = FALSE -- not blocklisted
           AND value['type'] = 'extension' -- nice webextensions only
+          AND value['foreign_install'] = FALSE  -- not sideloaded
           AND value['signed_state'] = 2 -- fully reviewed addons only
           AND value['user_disabled'] = FALSE -- active addons only get counted
           AND value['app_disabled'] = FALSE -- exclude compatibility disabled addons


crankycoder:
cc/ @wlach

mozetl/taar/taar_similarity.py
```diff
@@ -36,26 +41,27 @@ def get_samples(spark):
...
+        "geo_city[0] as geo_city",
+        "subsession_length[0] as subsession_length",
+        "settings[0].locale as locale",
+        "os as os",
+        "places_bookmarks_count[0].sum AS bookmark_count",
+        "scalar_parent_browser_engagement_tab_open_event_count[0].value AS tab_open_count",
+        "scalar_parent_browser_engagement_total_uri_count[0].value AS total_uri",
+        "scalar_parent_browser_engagement_unique_domains_count[0].value AS unique_tlds",
     )
+    df.cache()
```
wlach: If we're going to put this in as-is, let's add a comment explaining why this was added as a workaround.

crankycoder:
I've minimized the changes to just the cache() invocation and some extra docstrings now.

acmiyaguchi:
@crankycoder good insight on finding how to reproduce the test failure. I'm not entirely sure what's going on with the temp views, but materializing the data seems to help. I think it's probably a good idea to move the caching into the tests instead of the adding it to the job. It might cause issues in production otherwise. 

Using `yield` lets us apply `cache` to the dataframe, run the test, and then `unpersist` as cleanup. This seems to also fix the intermittent tests. 

```diff
diff --git a/tests/test_taar_similarity.py b/tests/test_taar_similarity.py
index e8a8468..e650182 100644
--- a/tests/test_taar_similarity.py
+++ b/tests/test_taar_similarity.py
@@ -185,7 +185,10 @@ def multi_clusters_df(generate_data):
             sample_snippets.append(variation)
             counter = counter + 1
 
-    return generate_data(sample_snippets)
+    df = generate_data(sample_snippets)
+    df.cache()
+    yield df
+    df.unpersist()
 
 
 @pytest.fixture
@@ -243,7 +246,6 @@ def test_similarity():
     assert taar_similarity.similarity_function(test_user_1, test_user_5)
 
 
-@pytest.mark.skip(reason="test intermittently fails")
 def test_get_addons(spark, addon_whitelist, multi_clusters_df):
     multi_clusters_df.createOrReplaceTempView("longitudinal")
 
@@ -256,7 +258,6 @@ def test_get_addons(spark, addon_whitelist, multi_clusters_df):
     assert isinstance(addons_df.schema.fields[1].dataType, ArrayType)
 
 
-@pytest.mark.skip(reason="test intermittently fails")
 def test_compute_donors(spark, addon_whitelist, multi_clusters_df):
     multi_clusters_df.createOrReplaceTempView("longitudinal")
```

crankycoder:
@acmiyaguchi thanks anthony - i've made the modifications as per your suggestion.

PR #256, opened 2018-09-12
Make hardware report idempotent

fbertsch:
This will enable rerunning of historical hardware reports. This has been tested in telemetry-test-bucket.

At this point I'm not even sure who to tag as reviewer here... tag @acmiyaguchi

### DIFF
--- a/mozetl/hardware_report/hardware_dashboard.py
+++ b/mozetl/hardware_report/hardware_dashboard.py
@@ -50,16 +50,22 @@ def main(start_date, end_date, bucket):
     # Concat the json files into the output.
     print "Joining JSON files..."
 
-    read_files = glob.glob("*.json")
+    new_files = set(report.keys())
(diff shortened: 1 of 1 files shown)


fbertsch:
@acmiyaguchi You're right, we should have tests here. Unfortunately this is the spot that doesn't have any. This is blocking longitudinal so if this looks good by your eye, lets merge now and create a bug to add test coverage using `boto`. I would do it now but it looks like that will take awhile.

acmiyaguchi:
This test has some related patterns, such as calling the CLI runner and using moto to mock buckets. 
https://github.com/mozilla/python_mozetl/blob/4bd8ceed909ae779826c383cbf7659b5dbf5869b/tests/test_topline_dashboard.py#L107-L150

It's probably enough to compartmentalize the main function into smaller, testable pieces though. It'll make it easier to review and modify in the future.

PR #258, opened 2018-09-14
Bug 1491330 - allow different submission_dates

fbertsch:
Why we're getting multiple types here I don't know, but I'm much too fatigued by longitudinal to figure out why.


fbertsch:
Hmm, this seems to indicate that they have never included millis: https://sql.telemetry.mozilla.org/queries/58945/source

acmiyaguchi:
This is probably related to the change that I made to replace joda-time with java.time in mozilla/telemetry-batch-view#477. In particular, `DateTimeFormatter.ISO_OFFSET_DATE_TIME` is like the jodatime equivalent, but without the millis e.g. `2011-12-03T10:15:30+01:00'`. I didn't realize it would break jobs :/

Redash might be parsing the datestrings in the query and formatting it in a standard way.

fbertsch:
Thanks @acmiyaguchi! Good to know about the cause.

PR #260, opened 2018-09-21
Do not set logging level by default

harterrt:
This will allow us to use mozetl.search.aggregates from databricks.
Currently, the output is overwhelmed by logging messages

### DIFF
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -32,9 +32,6 @@ MAX_CLIENT_SEARCH_COUNT = 10000
 SEARCH_AGGREGATES_VERSION = 4
 SEARCH_CLIENTS_DAILY_VERSION = 4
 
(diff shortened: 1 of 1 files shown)


wlach:
If this fixes the annoying amount of logging output I'm seeing on databricks, I'm all in favor! Databricks has been awesome for backfilling, way faster and much cheaper.

harterrt:
That's the hope! @mreid-moz, how can we test this version of the mozetl library in databricks? Here's an example notebook that reproduces the issue: https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/34767/command/34771

wlach:
@harterrt I typically test things by copypasting into a databricks notebook. e.g. https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/32272/command/32273

Issue #236, opened 2018-06-21
Add ability to import a generic whitelist before model training.


mlopatka:
https://github.com/mozilla/python_mozetl/blob/0f8189f87f857f43e9c0142f9c612a0bcc28978c/mozetl/taar/taar_lite_guidguid.py#L76

All taar ETL jobs should accept a generic whitelist of addons *before* model training jobs.

mlopatka:
@crankycoder is this addressed by the work in #237 ?
If so can you close it out, otherwise, lets discuss (in thread) what is the best way forward here.

PR #262, opened 2018-10-10
Fixes: #236

birdsarah:
We no longer need to do this given new whitelisting approach


mlopatka:
@acmiyaguchi @Dexterp37 just comment cleanup, can we get an approved reviewer to merge this in please.

PR #264, opened 2018-10-23
reformatted code with black

crankycoder:
This is just a reformatting PR before making changes to the similarity job to use clients_daily instead of longitudinal


mlopatka:
@acmiyaguchi or @Dexterp37 
can we get a review pleaze... this is mostly just formatting cleanup.

PR #265, opened 2018-10-23
Port TAAR similarity job to use clients_daily

crankycoder:
WIP : 

Note - This depends on #264 which reformats the similarity code

This patch repoints the TAAR similarity job to use the clients_daily table instead of the longitudinal table.


tests/test_taar_similarity.py
```diff
@@ -147,40 +137,52 @@ def multi_clusters_df(generate_data):
             # Generate some valid addons for each client. Each stub
             # addon as a specific GUID that contains both the variation
             # id and the addon id for simplified debugging.
-            variation["active_addons"] = [{}]
+            variation["active_addons"] = []
             for addon_id in range(0, 3):
                 addon_name = "var-{}-guid-{}".format(variation_id, addon_id)
-                variation["active_addons"][0][addon_name] = {
```
crankycoder: cc @mlopatka 

I don't think this test case wasn't written correctly before.  The `active_addons` field is defined as : 

```
    StructField("active_addons",         ArrayType(
        MapType(StringType(), StructType([
            StructField("blocklisted",      BooleanType(), True),
            StructField("type",             StringType(), True),
            StructField("signed_state",     LongType(), True),
            StructField("user_disabled",    BooleanType(), True),
            StructField("app_disabled",     BooleanType(), True),
            StructField("is_system",        BooleanType(), True),
            StructField("foreign_install",  BooleanType(), True)
        ]), True), True)),
```

which is roughly "list of dictionaries containing a dict that is `{addon_guid: {addon_metadata_dict}}`.

The loop on line 151 (was 141) now appends the dictionaries to the active_addons list.

mozetl/taar/taar_similarity.py
```diff
@@ -65,20 +82,33 @@ def get_addons_per_client(users_df, addon_whitelist, minimum_addons_count):
...
+
+    def flatten_valid_guid_generator(p):
+        for addon_data in p["active_addons"]:
+            for guid, data in addon_data.items():
+                if not is_valid_addon(guid, data):
+                    continue
+                yield guid
+
     return (
-        users_df.rdd
```
crankycoder: cc/ @mlopatka  

Based on the test changes in https://github.com/mozilla/python_mozetl/pull/265/files#r227487130 , I reworked the way that valid addon GUIDs are flattened into a simple list.

crankycoder: For reference - the longitudinal table had an array of struct's in the `active_addons` field - probably to denote the set of addons installed over time.

crankycoder:
I'm waiting on the clients_daily table PR in https://github.com/mozilla/telemetry-batch-view/pull/490 to go live so I can test some snippets in databricks and verify that my changes to the test suite are sane.

tests/test_taar_similarity.py
crankycoder: Bah.  Ignore https://github.com/mozilla/python_mozetl/pull/265#discussion_r227487130

On testing with live data on analysis.telemetry.mozilla.org, the active_addons column is a list of dictionaries of metadata.  There is no intermediary MapType involved.

crankycoder:
I've revised the testcase one last time to reflect what I've tested on an analysis.telemetry.m.o cluster with real data in the clients_daily table.

crankycoder:
ping @acmiyaguchi

PR #266, opened 2018-10-24
Fix some flake8 issues which were causing the build to fail

wlach:


wlach:
I think the unit tests cover this already, but fwiw here's the python console log of me verifying the assumptions I mentioned in the commit.

```
>>> df = s.createDataFrame([('65.5.3',)], ['s',])
>>> df.select(split(df.s, '\\.').alias('s')).collect()
[Row(s=[u'65', u'5', u'3'])]                                                    
>>> df.select(split(df.s, '\.').alias('s')).collect()
[Row(s=[u'65', u'5', u'3'])]
```

PR #269, opened 2018-11-26
Ported taar_locale to use clients_daily instead of using longitudinal table

crankycoder:
This migrates the taar_locale job to use the clients_daily table from the deprecated longitudinal table.

### DIFF
--- a/mozetl/taar/taar_locale.py
+++ b/mozetl/taar/taar_locale.py
@@ -36,7 +36,7 @@ def get_addons(spark):
     get_addons function.
     """
     return spark.sql("""
-        WITH sample AS
+       WITH sample AS
        (
           SELECT
             client_id,
@@ -64,7 +64,6 @@ def get_addons(spark):
             AND locality <> 'null'
             AND col.addon_id is not null
         ),
-
         country_addon_pairs AS (
         SELECT
         COUNT(*) AS pair_cnts, addon_key, locality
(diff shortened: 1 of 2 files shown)


crankycoder:
@mlopatka  r?

crankycoder:
@acmiyaguchi the test failure looks like it's not related to the code in this patch.    It looks like boto is trying to access S3 credentials for some reason.

mreid-moz:
@crankycoder you should be able to rebase on master (after #271) to fix this test failure.

mozetl/taar/taar_locale.py
```diff
@@ -37,26 +36,33 @@ def get_addons(spark):
...
     return spark.sql("""
-        WITH sample AS (
-        SELECT client_id,
-        settings[0].locale AS locality,
-        EXPLODE(active_addons[0])
-        FROM longitudinal
-        WHERE normalized_channel='release'
-          AND build IS NOT NULL
-          AND build[0].application_name='Firefox'
+        WITH sample AS
```
Dexterp37: nit: this should probably be indented with the lower `(`

crankycoder: Doh. I've pushed a commit to clean up that whitespace.

mreid-moz:
Let's hold off merging this until next week due to the allhands code freeze.

PR #279, opened 2018-12-17
Add normalized install frequencies to taar_locale. 

mlopatka:


mlopatka:
@acmiyaguchi @Dexterp37 @crankycoder 
I'm done fighting with CI, Anyone mind reviewing this?

mlopatka:
@acmiyaguchi `loc_norm` is the total number of all addon installations in a given locale. So locales with a larger number of clients or a particularly high addon engagement rate will both have a high value for loc_norm. The normalized frequency that is added in this PR is normalized such that the "popularity" of addons in the top10 are scaled by their relative popularitty against all addons for clients in that locale.

PR #282, opened 2018-12-20
Add Python 3 support 

### LINKED ISSUE #273: Python 3 support
This repository currently only supports python 2 packages. This should be running on python 3 by default.

acmiyaguchi:
This fixes #273, allowing for python 3 in the project.

* Update CI to include py27/py36 in tox and circleci
* Format the code base with black by fixing the minimum amount of syntax incompatibility via 2to3
* Format using 2to3 + black
* Individually fix each of the broken tests in CI
* Add black as one of the linting commands

I tried to make the manual changes as minimal as possible. I still need to run the jobs within EMR/Databricks to ensure that jobs are running correctly.

Finally, there's probably some need to add better dependency management and packaging. I've looked at some options:

* pip-tools and requirements.txt
* pipenv and Pipfile.lock
* poetry and pyproject.toml

They all share a common weakness of not supporting both major versions of Python. It seems best to wean off py27 first and then pin dependencies to py36.


mozetl/taar/taar_amodump.py
```diff
@@ -247,16 +247,16 @@ def marshal(value, name, type_def):
     serializers = {
         typing.List: list,
         typing.Dict: dict,
-        str: unicode,
-        unicode: unicode,
+        str: str,
+        str: str,
```
acmiyaguchi: Unicode is handled differently in python 3, so marshaling unicode isn't valid anymore. @crankycoder do you think this would affect the AMO dump, given that the tests pass in both version?

mozetl/utils.py
```diff
@@ -48,20 +48,15 @@ def write_csv(dataframe, path, header=True):
     is ~15GB for ec2 c3.xlarge (due to caching overhead).
     """
 
-    # NOTE: Before spark 2.1, toLocalIterator will timeout on some dataframes
-    # because rdd materialization can take a long time. Instead of using
-    # an iterator over all partitions, collect everything into driver memory.
-    logger.info("Writing {} rows to {}".format(dataframe.count(), path))
+    folder = "{}.temp".format(path)
+    logger.info("Writing single partition to {}".format(folder))
+    dataframe.repartition(1).write.csv("file://" + folder, header=header)
```
acmiyaguchi: I wanted to prevent version specific versions of the `csv.writer` from affecting the results with respect to unicode values. Using the spark dataframe writer avoids the python standard library and uses Spark's internal representation.

mozetl/maudau/maudau.py
```diff
@@ -149,7 +149,7 @@ def write_locally(results):
     basename = MAUDAU_SNAPSHOT_TEMPLATE.format(publication_date)
     cols = ["day", "dau", "mau", "generated_on"]
     with open(basename, "w") as f:
-        writer = csv.DictWriter(f, cols)
+        writer = csv.DictWriter(f, cols, lineterminator="\n")
```
acmiyaguchi: The carriage return isn't added via Spark, so I've changed the behavior here too. The previous roll-up is read and written over here, so there shouldn't be any conflicts.

setup.py
```diff
@@ -35,7 +35,8 @@
         'requests-toolbelt==0.8.0',
         'requests==2.20.1',
         'scipy==1.0.0rc1',
-        'typing==3.6.4'
+        'typing==3.6.4',
+        'six',
```
acmiyaguchi: This should eventually be pinned.

mozetl/testpilot/pulse.py
```diff
@@ -31,14 +31,13 @@ class Request(object):
...
-        [
-            # For some reason, field_types needs to be sorted. Use
-            # PYTHONHASHSEED = 3201792604 to reproduce failure
-            StructField(key, field_types[key][1], True)
-            for key in sorted(field_types)
-        ]
-    )
+    # python3 work-around to keep field_types in scope since class variables
+    # are not in the local scope of list comprehensions.
+    _create_struct = lambda x: StructField(x[0], x[1][1], True)  # noqa
```
acmiyaguchi: Definitions in the class scope require the instance or class reference.

tests/test_pulse.py
```diff
@@ -92,4 +92,4 @@ def test_simple_transform(row_to_dict, simple_rdd, spark_context):
         "trackingProtection": False,
     }
 
-    assert row_to_dict(actual) == expected
+    assert row_to_dict(actual, recursive=False) == expected
```
acmiyaguchi: This test has nested rows that shouldn't be recursively jsonified.

relud:
This should be split into more PRs. 4.5K lines isn't an amount I can confidently review in a reasonable amount of time.

I think it would be best to start with the smallest and least programmatic changes, and then work towards the largest ones (like adding black)

acmiyaguchi:
I think the best way to review the changes are reading the commits from top to bottom and navigating between them. I've organized it in the following way:

1. 6ffc0a5 - 40036a4 : modifications to CI
2. 3808db0  - 049728f: black + 2to3
3. f121af2 - db08f0d: changes required to fix tests from 2to3 process
4. dd20959: helper

If you ignore 6f4a8fb for formatting with black, the individual commits have much smaller (and reviewable) change-sets. I'd like to have black earlier in the process as opposed to later, so I can run it after each commit in the required fixes. 

I'll split this up into separate PRs, but I'll leave this here for reference.

relud:
> I'd like to have black earlier in the process as opposed to later, so I can run it after each commit in the required fixes.

okay, just be aware that black is the change that will take me the longest to review by far. 4K lines is going to take me days to find the time just to read it. Everything else combined I'm comfortable getting reviewed in <1 day.

acmiyaguchi:
No problem, I'm comfortable editing history so it doesn't matter if formatting comes first or last. I cherry-picked the changes onto a clean 2to3 branch w/o issue.

PR #290, opened 2018-12-21
[part 1] Add python3 and black to CI

acmiyaguchi:
This PR is the first part of a series to add python 3 support to mozetl. This adds CI support for py27, py36, and black (using the `--check` option). 

See #282 for the final state.

PR #291, opened 2018-12-21
[part 2] Run 2to3 and fix issues

acmiyaguchi:
This is part 2 of adding python 3 support in mozetl. This fixes broken tests by 2to3 and general incompatibilities between python 2 and 3.

See #290 for part 1 that adds CI
See #282 for the final state in this series of PRs

### DIFF
--- a/mozetl/engagement/churn/job.py
+++ b/mozetl/engagement/churn/job.py
@@ -36,6 +36,7 @@ from pyspark.sql.window import Window
 from . import release, utils
 from .schema import churn_schema
 from .utils import DS, DS_NODASH
+from functools import reduce
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -375,7 +376,7 @@ def clean_columns(prepared_clients, effective_version, start_ds):
     # Set the attributes to null if it's invalid
     select_attr = utils.build_col_expr({
         attr: F.when(F.col(is_valid), expr).otherwise(F.lit(None))
-        for attr, expr in utils.preprocess_col_expr(attr_mapping).iteritems()
+        for attr, expr in utils.preprocess_col_expr(attr_mapping).items()
     })
     select_metrics = utils.build_col_expr(metric_mapping)
     select_expr = select_attr + select_metrics
@@ -456,7 +457,7 @@ def transform(main_summary, effective_version, start_ds):
     records = (
         cleaned_data
         .groupBy(*attributes)
-        .agg(*agg_expr.values())
+        .agg(*list(agg_expr.values()))
     )
 
     return records.select(columns)
--- a/mozetl/engagement/retention/job.py
+++ b/mozetl/engagement/retention/job.py
@@ -43,6 +43,7 @@ from mozetl.engagement.churn.job import (
     MAX_SUBSESSION_LENGTH
 )
 from .schema import retention_schema
+from functools import reduce
 
 
 def valid_pcd(pcd, client_date):
--- a/mozetl/main.py
+++ b/mozetl/main.py
@@ -41,10 +41,10 @@ def etl_job(sc, sqlContext):
     results = transform_pings(get_data(sc))
 
     # Display results:
-    total = sum(map(operator.itemgetter(1), results.iteritems()))
+    total = sum(map(operator.itemgetter(1), iter(results.items())))
     # Print the OS and client_id counts in descending order:
     sorted_results = sorted(
-        results.iteritems(),
+        iter(results.items()),
         key=operator.itemgetter(1),
         reverse=True,
     )
--- a/mozetl/taar/taar_amodump.py
+++ b/mozetl/taar/taar_amodump.py
@@ -6,6 +6,7 @@ import logging
 import logging.config
 import typing
 from six.moves import urllib
+from six import text_type
 import queue as queue
 
 from .taar_utils import store_json_to_s3
@@ -244,8 +245,8 @@ class Undefined:
 def marshal(value, name, type_def):
     serializers = {typing.List: list,
                    typing.Dict: dict,
-                   str: unicode,
-                   unicode: unicode,
+                   str: text_type,
+                   text_type: text_type,
                    int: int,
                    float: float,
                    bool: bool}
--- a/mozetl/taar/taar_amodump.py
+++ b/mozetl/taar/taar_amodump.py
@@ -5,7 +5,7 @@ import json
 import logging
 import logging.config
 import typing
-import urllib.request, urllib.parse, urllib.error
+from six.moves import urllib
 import queue as queue
 
 from .taar_utils import store_json_to_s3
--- a/mozetl/taar/taar_amodump.py
+++ b/mozetl/taar/taar_amodump.py
@@ -5,9 +5,8 @@ import json
 import logging
 import logging.config
 import typing
-from six.moves import urllib
+from six.moves import urllib, queue
 from six import text_type
-import queue as queue
 
 from .taar_utils import store_json_to_s3
 
--- a/mozetl/taar/taar_amodump.py
+++ b/mozetl/taar/taar_amodump.py
@@ -5,10 +5,10 @@ import json
 import logging
 import logging.config
 import typing
-import urllib
-import Queue as queue
+import urllib.request, urllib.parse, urllib.error
+import queue as queue
 
-from taar_utils import store_json_to_s3
+from .taar_utils import store_json_to_s3
 
 import requests
 from requests_toolbelt.threaded import pool
@@ -93,7 +93,7 @@ class AMODatabase:
         addon_map = self._fetch_pages()
         self._fetch_versions(addon_map)
         final_result = {}
-        for k, v in addon_map.items():
+        for k, v in list(addon_map.items()):
             if 'first_create_date' in v:
                 final_result[k] = v
         logger.info("Final addon set includes %d addons." % len(final_result))
@@ -127,7 +127,7 @@ class AMODatabase:
         logger.info("Filling initial verson page queue")
 
         def iterFactory(guid_map):
-            for guid in guid_map.keys():
+            for guid in list(guid_map.keys()):
                 yield "https://addons.mozilla.org/api/v3/addons/addon/%s/versions/" % guid
 
         def chunker(seq, size):
@@ -187,7 +187,7 @@ class AMODatabase:
                     results = jdata['results']
 
                     raw_guid = resp.url.split("addon/")[1].split("/versions")[0]
-                    guid = urllib.unquote(raw_guid)
+                    guid = urllib.parse.unquote(raw_guid)
                     create_date = results[-1]['files'][0]['created']
 
                     record = addon_map.get(guid, None)
@@ -252,7 +252,7 @@ def marshal(value, name, type_def):
 
     if issubclass(type_def, JSONSchema):
         obj = {}
-        for attr_name, attr_type_def in type_def.meta.items():
+        for attr_name, attr_type_def in list(type_def.meta.items()):
             attr_value = value.get(attr_name, Undefined)
             if attr_value is not Undefined:
                 # Try marshalling the value
@@ -270,7 +270,7 @@ def marshal(value, name, type_def):
             k_cast, v_cast = type_def.__args__
             dict_vals = [(marshal(k, name, k_cast),
                           marshal(v, name, v_cast))
-                         for k, v in value.items()]
+                         for k, v in list(value.items())]
             return dict(dict_vals)
     else:
         return serializers[type_def](value)
--- a/mozetl/testpilot/pulse.py
+++ b/mozetl/testpilot/pulse.py
@@ -36,13 +36,13 @@ class Request(object):
     def __init__(self, request_dict):
         args = {
             field: conversion(request_dict.get(field))
-            for field, (conversion, sql_type) in Request.field_types.iteritems()
+            for field, (conversion, sql_type) in Request.field_types.items()
         }
         self.row = Row(**args)
 
 
 def _requests_to_rows(requests):
-    out = {k: Request(v).row for k, v in requests.iteritems()}
+    out = {k: Request(v).row for k, v in requests.items()}
     return out
 
 
--- a/mozetl/testpilot/pulse.py
+++ b/mozetl/testpilot/pulse.py
@@ -25,11 +25,10 @@ class Request(object):
 
     # python3 work-around to keep field_types in scope since class variables
     # are not in the local scope of list comprehensions.
-    _create_struct = lambda x: StructField(x[0], x[1][1], True)  # noqa
     _keys = sorted(field_types.keys())
     _dtypes = map(field_types.get, _keys)
 
-    StructType = StructType(list(map(_create_struct, zip(_keys, _dtypes))))
+    StructType = StructType([StructField(x[0], x[1][1], True) for x in zip(_keys, _dtypes)])
 
     def __init__(self, request_dict):
         args = {
--- a/mozetl/testpilot/pulse.py
+++ b/mozetl/testpilot/pulse.py
@@ -23,15 +23,13 @@ class Request(object):
         'time': int_type,
     }
 
-    StructType = StructType([
-        # For some reason, field_types needs to be sorted. Use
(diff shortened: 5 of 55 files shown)


mozetl/topline/topline_summary.py
```diff
@@ -214,7 +215,7 @@ def transform(dataframe, start, mode):
         .join(searches, attributes, "outer")
         .join(hours, attributes, "outer")
         .withColumnRenamed('country', 'geo')
-        .withColumn('crashes', F.lit(0L))
+        .withColumn('crashes', F.lit(0))
```
relud: please double-check that this creates a column of type long in both python 2 and python 3

mozetl/utils.py
```diff
@@ -48,20 +48,15 @@ def write_csv(dataframe, path, header=True):
...
-            writer.writerow(dataframe.columns)
-
-        for row in dataframe.collect():
-            row = [unicode(s).encode('utf-8') for s in row]
-            writer.writerow(row)
+    # there be a single `part-*`` file in the folder ending with csv
+    csv_file = glob.glob("{}/*.csv".format(folder))[0]
+    logger.info("Moving {} to {}".format(csv_file, path))
+    shutil.copy(csv_file, path)
+    shutil.rmtree(folder)
```
relud: could you explain this change?

setup.py
```diff
@@ -35,7 +35,8 @@
         'requests-toolbelt==0.8.0',
         'requests==2.20.1',
         'scipy==1.0.0rc1',
-        'typing==3.6.4'
+        'typing==3.6.4',
+        'six',
```
relud: please add a six version or a comment in the file explaining why it doesn't have a version

tests/test_maudau.py
```diff
@@ -27,7 +27,7 @@ def make_frame(spark):
         ('a', '2017-05-05', '20170507'),
     ]
     return spark.createDataFrame(
-        [dict(zip(cols, tup)) for tup in values],
+        [dict(list(zip(cols, tup))) for tup in values],
```
relud: this `list` is unnecessary, the `dict` works just fine in python 2 and 3

tests/test_maudau.py
```diff
@@ -101,7 +101,7 @@ def test_write_locally():
         {'day': '20170504', 'dau': 1, 'mau': 3,
          'generated_on': generated}
     ]
-    expected = '\r\n'.join([
```
relud: why are you removing carriage returns here?

tests/test_landfill_sampler.py
```diff
@@ -11,19 +11,19 @@ def sample_document():
...
-            'Timestamp': 1534691279765301222L,
-            'Type': u'telemetry-raw',
-            u'User-Agent': u'pingsender/1.0',
-            u'X-Forwarded-For': u'127.0.0.1',
-            u'X-PingSender-Version': u'1.0',
-            u'args': u'v=4',
-            u'protocol': u'HTTP/1.1',
-            u'remote_addr': u'1.1.1.1',
-            u'uri': u'/submit/telemetry/doc-id/main/Firefox/61.0.2/release/20180807170231'
+            'Content-Length': '7094',
```
relud: removing the `u` is a functional change here for python 2, and a noop in python 3. why is it being changed?

tests/test_pulse.py
```diff
@@ -81,4 +81,4 @@ def test_simple_transform(row_to_dict, simple_rdd, spark_context):
         'trackingProtection': False,
     }
 
-    assert row_to_dict(actual) == expected
+    assert row_to_dict(actual, recursive=False) == expected
```
relud: why is recursive false?

tests/test_taar_amowhitelist.py
```diff
@@ -269,13 +270,13 @@
...
             "zh-CN":
-                "\\u4f7f\\u7528OSD Lyrics\\u663e\\u793a\\u8c46\\u74e3\\u7535\\u53f0\\u6b4c\\u8bcd"
+                "\u4f7f\u7528OSD Lyrics\u663e\u793a\u8c46\u74e3\u7535\u53f0\u6b4c\u8bcd"
         },
         "tags": [
             "douban",
             "osd-lyrics",
             "osdlyrics",
-            "\\u8c46\\u74e3"
+            "\u8c46\u74e3"
```
relud: with a `u"` or a `"...\\u..."` these statements behave the same in python 2 vs 3, and with this they behave differently. if we're sure we want this, please add a comment in this file indicating why it's okay that this behaves differently in python 2 and 3.

tests/test_utils.py
```diff
@@ -58,8 +58,9 @@ def test_generate_filter_parameters():
     assert expected28 == actual28
 
 
+@pytest.mark.skip(reason="Handled correctly in python 3")
 def test_write_csv_valid_unicode(generate_data, tmpdir):
-    test_data = [u'∆', u'∫', u'∬']
+    test_data = ['∆', '∫', '∬']
```
relud: i think we should leave this in place, with the `u'`s, to ensure we still handle unicode correctly in python 2.7

tests/test_topline_summary.py
```diff
@@ -38,8 +38,8 @@ def generate_dates(submission_date_s3, ts_offset=0, creation_offset=0):
 
 def search_row(engine='hooli', count=1, source='searchbar'):
     return Row(
-        engine=unicode(engine),
-        source=unicode(source),
+        engine=str(engine),
+        source=str(source),
```
relud: this changes the test in python 2. I think we should use `six.text_type` here.

tests/test_utils.py
```diff
@@ -29,10 +29,10 @@ def test_write_csv_ascii(generate_data, tmpdir):
     path = str(tmpdir.join('test_data.csv'))
     utils.write_csv(df, path)
 
-    with open(path, 'rb') as f:
+    with open(path, "r") as f:
```
relud:
```suggestion
    with open(path, "rb") as f:
```

tests/test_utils.py
```diff
@@ -29,10 +29,10 @@ def test_write_csv_ascii(generate_data, tmpdir):
...
         data = f.read()
 
-    assert data.rstrip().split('\r\n')[1:] == test_data
+    assert data.rstrip().split("\n")[1:] == test_data
```
relud:
```suggestion
    assert [l.encode() for l in data.rstrip().split(b"\r\n")[1:]] == test_data
```

tests/test_utils.py
```diff
@@ -68,7 +69,7 @@ def test_write_csv_valid_unicode(generate_data, tmpdir):
     with open(path, 'rb') as f:
         data = f.read().decode('utf-8')
 
-    assert data.rstrip().split('\r\n')[1:] == test_data
+    assert data.rstrip().split("\n")[1:] == test_data
```
relud:
```suggestion
    assert [l.encode() for l in data.rstrip().split(b"\r\n")[1:]] == test_data
```

tests/test_utils.py
```diff
@@ -68,7 +69,7 @@ def test_write_csv_valid_unicode(generate_data, tmpdir):
     with open(path, 'rb') as f:
```
relud: I highly recommend searching to make sure all cases of `open(..., 'rb')` actually get accessed as bytes, using `b""` values and whatnot.

mozetl/engagement/churn/job.py
```diff
@@ -36,6 +36,7 @@
 from . import release, utils
 from .schema import churn_schema
 from .utils import DS, DS_NODASH
+from functools import reduce
```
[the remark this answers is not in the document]
acmiyaguchi: It gets used a couple times to join a list of where clauses. 

https://github.com/mozilla/python_mozetl/blob/6bb8b286cb56f0d952846b8bb193f14645e9b3fa/mozetl/engagement/churn/job.py#L177-L181

mozetl/engagement/retention/job.py
```diff
@@ -43,6 +43,7 @@
     MAX_SUBSESSION_LENGTH
 )
 from .schema import retention_schema
+from functools import reduce
```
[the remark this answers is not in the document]
acmiyaguchi: https://github.com/mozilla/python_mozetl/blob/6bb8b286cb56f0d952846b8bb193f14645e9b3fa/mozetl/engagement/retention/job.py#L53-L61

mozetl/main.py
```diff
@@ -41,17 +41,17 @@ def etl_job(sc, sqlContext):
...
     )
     for pair in sorted_results:
-        print(
+        print((
             "OS: {:<10} Percent: {:0.2f}%".format(
                 pair[0],
                 float(pair[1]) / total * 100
             )
-        )
+        ))
```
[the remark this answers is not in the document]
acmiyaguchi: It doesn't, but it's probably related to this behavior:

> When the -p is passed, 2to3 treats print as a function instead of a statement. This is useful when from __future__ import print_function is being used. If this option is not given, the print fixer will surround print calls in an extra set of parentheses because it cannot differentiate between the print statement with parentheses (such as print ("a" + "b" + "c")) and a true function call.

https://docs.python.org/3.0/library/2to3.html

I can see if I can minimize this, but it occurs frequently enough.

mozetl/taar/taar_amodump.py
```diff
@@ -5,10 +5,10 @@
 import logging
 import logging.config
 import typing
-import urllib
-import Queue as queue
+from six.moves import urllib
+import queue as queue
```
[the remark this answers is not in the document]
acmiyaguchi: Artifact of 2to3

tests/test_taar_amowhitelist.py
acmiyaguchi: Good catch, I'll drop this commit because I thought it had been added by 2to3.

tests/test_utils.py
acmiyaguchi: I think there is some incompatibility between 2 and 3 here in terms of IO behavior, but I'll take a second look.

mozetl/engagement/churn/job.py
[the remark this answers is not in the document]
relud: oh, i see now. in python 2 it's a keyword. tyty

mozetl/main.py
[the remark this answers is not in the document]
relud: To me that reads as adding `-p` should prevent 2to3 from doing this

tests/test_taar_amowhitelist.py
relud: separately we should fix cases where we use `"...\u..."` instead of `"...\\u.."` and double the backslash (unless its a `u"`)

tests/test_utils.py
relud: explanation: in python 3 if you don't read as binary then "\r\n" becomes "\n", but if you read as binary then decode it works the same in 2 and 3.

tests/test_utils.py
relud: also, I think we should use `[u'\u2206', u'\u222b', u'\u222c']` instead of using literal unicode characters in the file, but that's not really part of the python 3 work.

mozetl/testpilot/pulse.py
```diff
@@ -23,26 +23,24 @@ class Request(object):
...
-            True
-        ) for key in sorted(field_types)
-    ])
+    # python3 work-around to keep field_types in scope since class variables
+    # are not in the local scope of list comprehensions.
+    _create_struct = lambda x: StructField(x[0], x[1][1], True)  # noqa
+    _keys = sorted(field_types.keys())
+    _dtypes = map(field_types.get, _keys)
+
+    StructType = StructType(list(map(_create_struct, zip(_keys, _dtypes))))
```
relud: I think this would be better than using `# noqa` and naming a lambda function:
```suggestion
    StructType = StructType([StructField(x[0], x[1][1], True) for x in zip(_keys, _dtypes)])
```

acmiyaguchi: I'll try this out, maybe the `_keys` and `_dtypes` variables are enough to avoid the class scoping issues.

tests/test_utils.py
acmiyaguchi: That makes sense, I took the opposite approach to convert everything `\r\n` to `\n` to fix the issue.

tests/test_taar_amowhitelist.py
```diff
@@ -92,7 +92,7 @@
             "text_count": 1.0
         },
         "summary": {
-            "en-US": "Luffi - Let\u00b4s use freeware! fun included:)"
+            "en-US": "Luffi - Let\\u00b4s use freeware! fun included:)"
```
acmiyaguchi: This unicode change does get added in 2to3, and I end up reversing this.

mozetl/main.py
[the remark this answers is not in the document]
acmiyaguchi: There are some degenerate cases where adding `-p` won't convert the print statement correctly without adding the parenthesis (I ran into this before).

```
RefactoringTool: There were 5 errors:
RefactoringTool: Can't parse mozetl/addon_aggregates/addon_aggregates.py: ParseError: bad input: type=3, value=u'"loading..."', context=(u' ', (68, 10))
RefactoringTool: Can't parse mozetl/hardware_report/hardware_dashboard.py: ParseError: bad input: type=3, value=u'"Joining JSON files..."', context=(u' ', (51, 10))
RefactoringTool: Can't parse mozetl/tab_spinner/generate_counts.py: ParseError: bad input: type=3, value=u'"Start Date: {}, End Date: {}"', context=(u' ', (24, 10))
RefactoringTool: Can't parse mozetl/tab_spinner/utils.py: ParseError: bad input: type=3, value=u'"Short and longcounts match"', context=(u' ', (156, 14))
RefactoringTool: Can't parse tests/test_addon_aggregates.py: ParseError: bad input: type=1, value=u'first_install', context=(u' ', (151, 14))
```
I can manually fix these up to reduce the noise though.

tests/test_taar_amowhitelist.py
relud: to be clear, using a `\\` correctly does the same thing in python 2 and 3, where reducing it to `\u` behaves differently (unless it used `u"`, then it behaves the same for both)

tests/test_maudau.py
acmiyaguchi: This is an artifact of the 2to3 conversion, it seems to be liberal in sprinkling lists around all of the iterators.

tests/test_pulse.py
acmiyaguchi: This line uses a `Row` object, instead of the json representation.

https://github.com/mozilla/python_mozetl/blob/d92d98833482170cab672aa922f380e1e79ae527/tests/test_pulse.py#L76

tests/test_taar_amowhitelist.py
acmiyaguchi: I'm not totally clear on this behavior, but I've left it as is.

acmiyaguchi:
There are a couple of comments that I still need to get to regarding the int/long behavior. I also want to run a few of these jobs in a real environment (probably via Databricks since we have a drop-in replacement in Airflow and has a managed python environment).

acmiyaguchi:
I've run into some issues running python 3.6 in databricks because it's running ubuntu 16.04 LTS. The same could be said for EMR, although this could be avoided by using the anaconda distribution in the bootstrap.

mozetl/topline/topline_summary.py
acmiyaguchi: This was a really good catch, I added schema validation in a test.

acmiyaguchi:
## Search Clients Daily
Python 2
https://dbc-caf9527b-e073.cloud.databricks.com/#job/1068/run/1
image

Python 3
https://dbc-caf9527b-e073.cloud.databricks.com/#job/1070/run/1
image

Verified with: `aws s3 ls s3://telemetry-test-bucket/test_search/v4/ --recursive | grep _SUCCESS`

## Churn
Python 3
https://dbc-caf9527b-e073.cloud.databricks.com/#job/1069/run/1
image

Verified with: `aws s3 ls s3://telemetry-test-bucket/churn/v3/ --recursive | grep _SUCCESS`

acmiyaguchi:
Everything is ready to go and should be compatible with our current environment. Today may not be the best day to deploy with the migration of the composer instance and recent changes to this repo in the last day.

PR #294, opened 2018-12-31
Add Sphinx API docs

acmiyaguchi:
This adds auto-generated documentation for modules in this package for better reference.

For an example of what this looks like, see: https://acmiyaguchi.github.io/python_mozetl/

This is the [documentation generated for the `mozetl.engagement.churn` module](https://acmiyaguchi.github.io/python_mozetl/source/mozetl.engagement.churn.html?highlight=churn), which looks reasonable for being auto-generated.


acmiyaguchi:
I took the documentation workflow from mozilla/python_moztelemetry and implemented it here. I use tox for generating the api docs which is run via `tox -e docs`.

PR #297, opened 2019-01-08
fixup! Add util for safe session stopping on Databricks

akkomar:


acmiyaguchi:
It might be a good idea to add a small unit test for coverage, otherwise LGTM

PR #300, opened 2019-01-09
fixup! Add util for safe session stopping on Databricks

akkomar:
This fixes a fix for safe session stopping util (condition was reversed).


acmiyaguchi:
That was easy to miss.

PR #301, opened 2019-01-19
Rewrite TAAR locale job to report DP-protected frequency weights

### LINKED ISSUE #280: Base add-ons query for taar_locale should restrict to 1 row per client
https://github.com/mozilla/python_mozetl/blob/09a64456ff290599fd5d7b1892e5704c9582eb15/mozetl/taar/taar_locale.py#L41-L49

The base query in `get_addons()` filters by `channel` and `app_name` but otherwise includes all rows.

Since `clients_daily` has 1 row per clientID per submission date, this means that we are double-counting clients. Eg., if a client was active on 5 different dates, they would have 5 rows in the dataset and contribute up to 5 to the `pair_cnts` for each of their add-ons. This query should probably select the row with the most recent submission date for each clientID.

Also, since `clients_daily` ranges back to early 2016, the query pulls in add-on installs over that entire time period. It should probably be limited to recent activity by `submission_date_s3` (say, to the last 3 months) for efficiency and relevance.

dzeber:
Add-on recommendations and frequency weights generated by the TAAR add-on recommender have been revised to incorporate differential privacy protection. Frequency counts are now computed for every whitelisted add-on in every locale and perturbed using the Laplace mechanism. Details of the methodology and an evaluation of parameter choices are provided in [this notebook](https://metrics.mozilla.com/~dzeber/notebooks/taar/taar_locale_privacy.html).

One implementation change to note is that the full table of raw locale/add-on frequency counts is now dumped to local memory and processed from there, whereas all processing was previously done in Spark until the final recommendations were generated. This is done to simplify dealing with random number generation, and relies on the assumption that the table remains relatively small (it has `num locales * len(whitelist)` rows).

This PR also fixes #280 (which was not yet fixed for this job).

This is a WIP because it still doesn't include dynamic selection of per-locale add-on limits, a feature we had discussed. However, this will be a relatively minor parameter selection change and does not affect the functioning of the job. Currently, all locale add-on limits are set to 1. Another point we can discuss further if necessary is the selection of the _epsilon_ parameter, which is currently set to 0.4.

r? @mlopatka @crankycoder


tests/test_taar_locale.py
```diff
@@ -35,142 +36,481 @@
...
+def same_rows(rows_list_1, rows_list_2):
+    """Compare DF rows represented as lists of tuples ('records').
+
+    Checks that the lists contain the same tuples (possibly in different orders).
+    """
+    return sorted(rows_list_1) == sorted(rows_list_2)
+
+
+# =========================================================================== #
+
```
mlopatka: Please include an additional test that locale/add-on output weights sum to 1.0 at output.

Similar expected output as is used in `test_get_top_addons_by_locale(addon_count_neg):` could be used, but I'd like to see the additional sanity check that our relative weights are comparable across locales after the noisification, scaling, and removal operations are all performed.

mlopatka:
flagging @acmiyaguchi for python_mozetl review.

dzeber:
I've made the requested changes and rebased onto the latest master. This is now ready for final review.

The locale add-on limits are left as is for now, and I've filed an issue about optimizing them later.

dzeber:
Thanks! No rush, it can wait till after the weekend.

PR #303, opened 2019-01-24
Fix test pilot experiments fetch

akkomar:
This is a quick fix for fetching test pilot experiments list.
Test pilot was retired making this function throw exception, this should be revisited with a proper fix.


akkomar:
Caveat: while unblocking other jobs, this might make addon_aggregates inconsistent with previous runs.

PR #309, opened 2019-02-05
Add a minimal click command for the tab spinner etl job

wlach:
This will let us use `mozetl-submit` inside airflow, which is quite
a bit cleaner.


wlach:
If this lands and works with airflow's `mozetl-submit`, we should delete `scheduled/tab_spinner.py`, as it will no longer be needed.

PR #310, opened 2019-02-05
Fix tab-spinner click command

wlach:
We were passing the wrong type of spark object to the actual etl
code.

### DIFF
--- a/mozetl/tab_spinner/tab_spinner.py
+++ b/mozetl/tab_spinner/tab_spinner.py
@@ -6,7 +6,7 @@ from generate_counts import run_spinner_etl
 @click.command()
 def main():
     spark = SparkSession.builder.appName("long_tab_spinners").getOrCreate()
-    run_spinner_etl(spark)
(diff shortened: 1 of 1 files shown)


mozetl/tab_spinner/tab_spinner.py
```diff
@@ -1,12 +1,13 @@
 import click
-from pyspark.sql import SparkSession
+from pyspark import SparkConf, SparkContext
 from generate_counts import run_spinner_etl
 
 
 @click.command()
 def main():
-    spark = SparkSession.builder.appName("long_tab_spinners").getOrCreate()
-    run_spinner_etl(spark)
```
acmiyaguchi: It's preferable to use the SparkSession, but then use the `sparkContext` attribute.

http://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.SparkSession.sparkContext

 `run_spinner_etl(spark.sparkContext)`

PR #311, opened 2019-02-06
adding bgbb jobs and tests

wcbeard:
Moving jobs and test from bgbb_lib into here.


acmiyaguchi:
@wcbeard Can you rebase on master -- there are some dependencies that have been pinned that will likely fix test failures.

PR #312, opened 2019-02-07
Pin pandas to 0.23.4

wlach:
It seems like the implicit upgrade to pandas 0.24 broke the tab spinner severity job, due to
a compatibility mismatch between the executor and workers. Let's pin our pandas version
to the last known working version (0.23.4) and add a comment warning people about this
in case they want to add new things in the future (the dependencies were explicitly pinned before 
for this exact reason). See: https://bugzilla.mozilla.org/show_bug.cgi?id=1525022


wlach:
Build is failing due to a boto issue, investigating in #313

PR #313, opened 2019-02-07
Pin boto dependencies

wlach:
Seems like v1.9.99 has an issue importing S3Transfer


wlach:
Found the root cause: https://github.com/boto/s3transfer/issues/130

Let's just use the last known working versions while we wait for that to be sorted out.

PR #316, opened 2019-02-13
Update `mozetl-databricks.py` for running external mozetl-compatible modules

acmiyaguchi:
This updates the Databricks workflow for supporting external modules that follow the mozetl convention. The convention is as follows:

* The module uses the `click` library for defining a command-line interface
* All arguments use the `@click.option` decorator. Arguments that are required should use the `required=True` parameter.
  - `click` will expose command-line arguments via environment variables
* The module has a top-level `cli` submodule such that `from my_module import cli` is a valid command

A module that adheres to these basic conventions can be scheduled via Airflow using standard tooling.

### DIFF
--- a/bin/mozetl-databricks.py
+++ b/bin/mozetl-databricks.py
@@ -29,9 +29,7 @@ def generate_runner(module_name, instance, token):
     logging.debug(dedent(runner_data))
 
     request = {
-        "content": b64encode(dedent(runner_data)),
-        "format": "SOURCE",
-        "language": "PYTHON",
+        "contents": b64encode(dedent(runner_data)),
         "overwrite": True,
         "path": "/FileStore/airflow/{module}_runner.py".format(module=module_name),
     }
@@ -41,7 +39,7 @@ def generate_runner(module_name, instance, token):
         "Authorization": "Bearer {token}".format(token=token),
         "Content-Type": "application/json",
     }
-    conn.request("POST", "/api/2.0/workspace/import", json.dumps(request), headers)
+    conn.request("POST", "/api/2.0/dbfs/put", json.dumps(request), headers)
     resp = conn.getresponse()
     logging.info("status: {} reason: {}".format(resp.status, resp.reason))
     logging.info(resp.read())
@@ -61,7 +59,7 @@ def run_submit(args):
             },
         },
         "spark_python_task": {
-            "python_file": "s3://telemetry-airflow/steps/mozetl_runner.py",
+            "python_file": "dbfs:/FileStore/airflow/{module}_runner.py".format(args.module_name),
             "parameters": args.command,
         },
         "libraries": {
--- a/bin/mozetl-databricks.py
+++ b/bin/mozetl-databricks.py
@@ -1,12 +1,30 @@
 #!/usr/bin/env python2
 
 import argparse
-import httplib
 import os
 import json
 import logging
 from base64 import b64encode
 from textwrap import dedent
+
+try:
+    from urllib.request import urlopen, Request
+except ImportError:
+    from urllib2 import urlopen, Request
+
+
+def api_request(instance, route, data, token):
+    api_endpoint = "https://{instance}/{route}".format(
+        instance=instance, route=route.lstrip("/")
+    )
+    headers = {
+        "Authorization": "Bearer {token}".format(token=token),
+        "Content-Type": "application/json",
+    }
+    req = Request(api_endpoint, data=data.encode(), headers=headers)
+    resp = urlopen(req)
+    logging.info("status: {} info: {}".format(resp.getcode(), resp.info()))
+    return resp
 
 
 def generate_runner(module_name, instance, token):
@@ -29,21 +47,13 @@ def generate_runner(module_name, instance, token):
     logging.debug(dedent(runner_data))
 
     request = {
-        "contents": b64encode(dedent(runner_data)),
+        "contents": b64encode(dedent(runner_data).encode()).decode(),
         "overwrite": True,
         "path": "/FileStore/airflow/{module}_runner.py".format(module=module_name),
     }
     logging.debug(json.dumps(request, indent=2))
-    conn = httplib.HTTPSConnection(instance)
-    headers = {
-        "Authorization": "Bearer {token}".format(token=token),
-        "Content-Type": "application/json",
-    }
-    conn.request("POST", "/api/2.0/dbfs/put", json.dumps(request), headers)
-    resp = conn.getresponse()
-    logging.info("status: {} reason: {}".format(resp.status, resp.reason))
+    resp = api_request(instance, "/api/2.0/dbfs/put", json.dumps(request), token)
     logging.info(resp.read())
-    resp.close()
 
 
 def run_submit(args):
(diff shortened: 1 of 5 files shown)


bin/mozetl-databricks.py
```diff
@@ -5,9 +5,48 @@
...
+        pass
+    """.format(
+        module=module_name
+    )
+    logging.debug(dedent(runner_data))
+
+    request = {
+        "contents": b64encode(dedent(runner_data)),
+        "overwrite": True,
+        "path": "/FileStore/airflow/{module}_runner.py".format(module=module_name),
```
wcbeard: This will require a keyword argument (`.format(module=module_name)`)

bin/mozetl-databricks.py
```diff
@@ -20,7 +59,7 @@ def run_submit(args, instance="dbc-caf9527b-e073.cloud.databricks.com"):
             },
         },
         "spark_python_task": {
-            "python_file": "s3://telemetry-airflow/steps/mozetl_runner.py",
+            "python_file": "dbfs:/FileStore/airflow/{module}_runner.py".format(args.module_name),
```
wcbeard: This will require a kw argument (`.format(module=args.module_name)`)

bin/mozetl-databricks.py
```diff
@@ -5,9 +5,48 @@
...
+        "Content-Type": "application/json",
+    }
+    conn.request("POST", "/api/2.0/dbfs/put", json.dumps(request), headers)
+    resp = conn.getresponse()
+    logging.info("status: {} reason: {}".format(resp.status, resp.reason))
+    logging.info(resp.read())
+    resp.close()
+
+
+def run_submit(args):
```
wcbeard: I think the `conn = httplib.HTTPSConnection(instance)` below now needs to be `conn = httplib.HTTPSConnection(args.instance)`?

bin/mozetl-databricks.py
```diff
@@ -5,9 +5,48 @@
...
+        # avoid calling sys.exit() in databricks
+        # http://click.palletsprojects.com/en/7.x/api/?highlight=auto_envvar_prefix#click.BaseCommand.main
+        pass
+    """.format(
+        module=module_name
+    )
+    logging.debug(dedent(runner_data))
+
+    request = {
+        "contents": b64encode(dedent(runner_data)),
```
wcbeard: I think this should be `.decode('ascii')`-ed, so that json.dumps works below (it doesn't take bytes).

Also, in python 3 at least, I need to `.encode('utf-8')` before passing to `b64encode` (it doesn't want a string). So all in all, `b64encode(dedent(runner_data).encode('utf-8')).decode('ascii')`

acmiyaguchi: The script is running via python 2 (for compat reasons), so this isn't an issue there.

acmiyaguchi:
I've updated the script with a few of the issues found during review, this is the log for a run of the `system_check` module.

```bash
$ python bin/mozetl-databricks.py \
>     --git-path https://github.com/acmiyaguchi/python_mozetl.git \
>     --git-branch federated-mozetl \
>     --token <TOKEN> \
>     system_check
DEBUG:root:
# This runner has been auto-generated from mozilla/python_mozetl/bin/mozetl-databricks.py.
# Any changes made to the runner file will be over-written on subsequent runs.
from mozetl import cli

try:
    cli.entry_point(auto_envvar_prefix="MOZETL")
except SystemExit:
    # avoid calling sys.exit() in databricks
    # http://click.palletsprojects.com/en/7.x/api/?highlight=auto_envvar_prefix#click.BaseCommand.main
    pass

DEBUG:root:{
  "path": "/FileStore/airflow/mozetl_runner.py", 
  "overwrite": true, 
  "contents": "CiMgVGhpcyBydW5uZXIgaGFzIGJlZW4gYXV0by1nZW5lcmF0ZWQgZnJvbSBtb3ppbGxhL3B5dGhvbl9tb3pldGwvYmluL21vemV0bC1kYXRhYnJpY2tzLnB5LgojIEFueSBjaGFuZ2VzIG1hZGUgdG8gdGhlIHJ1bm5lciBmaWxlIHdpbGwgYmUgb3Zlci13cml0dGVuIG9uIHN1YnNlcXVlbnQgcnVucy4KZnJvbSBtb3pldGwgaW1wb3J0IGNsaQoKdHJ5OgogICAgY2xpLmVudHJ5X3BvaW50KGF1dG9fZW52dmFyX3ByZWZpeD0iTU9aRVRMIikKZXhjZXB0IFN5c3RlbUV4aXQ6CiAgICAjIGF2b2lkIGNhbGxpbmcgc3lzLmV4aXQoKSBpbiBkYXRhYnJpY2tzCiAgICAjIGh0dHA6Ly9jbGljay5wYWxsZXRzcHJvamVjdHMuY29tL2VuLzcueC9hcGkvP2hpZ2hsaWdodD1hdXRvX2VudnZhcl9wcmVmaXgjY2xpY2suQmFzZUNvbW1hbmQubWFpbgogICAgcGFzcwo="
}
INFO:root:status: 200 reason: OK
INFO:root:{}

DEBUG:root:{
  "libraries": {
    "pypi": {
      "package": "git+https://github.com/acmiyaguchi/python_mozetl.git@federated-mozetl"
    }
  }, 
  "new_cluster": {
    "node_type_id": "c3.4xlarge", 
    "spark_version": "4.3.x-scala2.11", 
    "aws_attributes": {
      "instance_profile_arn": "arn:aws:iam::144996185633:instance-profile/databricks-ec2", 
      "availability": "ON_DEMAND"
    }, 
    "num_workers": 2
  }, 
  "spark_python_task": {
    "python_file": "dbfs:/FileStore/airflow/mozetl_runner.py", 
    "parameters": [
      "system_check"
    ]
  }, 
  "run_name": "mozetl local submission"
}
INFO:root:status: 200 reason: OK
INFO:root:{"run_id":6141}
```

bin/mozetl-databricks.py
wlach: It might be nice to make this forward compatible with python 3, if it's easy to do.

acmiyaguchi:
The dependency management issues with Spark/Pandas in the last few weeks have shown that managing dependencies in a mono repo can be difficult. I'm concerned about pulling in Python3-only libraries into the repository which is troublesome because of existing Python2 support. This PR is targeting support for the scripts in [bgbb_airflow](https://github.com/wcbeard/bgbb_airflow/), which itself uses new dependencies like numba and an up-to-date pandas that could cause some issues here because of the tight coupling of jobs (although the test suite does a fairly good job at catching a lot of errors). 

I wouldn't recommend starting a new repo as the default mode, because it's much simpler to follow the conventions here for writing a new job, testing it, and deploying it. However, I think it's good to have it as an option, especially for a project that's pulling very large dependencies. Regardless of how the jobs are managed, our Airflow repository is the gatekeeper for production-ready jobs. Jobs that are effectively ported notebooks should be scrutinized if they aren't part of mozetl or tbv.

wlach:
Ok, that makes sense! I'm ok with landing as long as people aren't scheduling jobs outside of python_mozetl as a way to do an end-run around proper review.

acmiyaguchi:
The `mozetl-databricks.py` script should be python 2/3 compatible now. I've tested it against the `system_check` job with success.

I also updated the README with some pointers on setting the `git-path` and `module-name` options for testing, along with a link to this PR.

acmiyaguchi:
@wlach Thanks for the review!

PR #318, opened 2019-02-21
Bug 1528363 - Only consider last 180 days (=6 months) for tab spinner job

wlach:
We actually were eliminating these old pings from consideration already,
but not in an efficient way.

### DIFF
--- a/mozetl/tab_spinner/generate_counts.py
+++ b/mozetl/tab_spinner/generate_counts.py
@@ -1,65 +1,43 @@
-import ujson as json
 import boto3
-
+import json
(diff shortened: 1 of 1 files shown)


wlach:
I tested this out on a databricks notebook on the shared serverless cluster, and the job ran in about 3 hours.

acmiyaguchi:
The code simplified nicely from not having to deal with the old-infra telemetry. Though one thing to note is that based on our data retention policy on our raw store, it doesn't seem like anything from 2016 should be queried in the job.

PR #319, opened 2019-02-25
Fix tab_spinner job

acmiyaguchi:
This fixes two issues with the tab_spinner job. The first is a relative import error in the module. The second is a serialization error due to the existence of the `numpy.float64` type.

The first is referenced in mozilla/telemetry-airflow#441. I've added a test that should prevent import errors from occurring again in `test_cli.py`. To verify that the tab spinner job works, I've tested the entire command using the `click.CliRunner`. This also caught the import error, as well as a hidden serialization error:

```
E       AssertionError: assert -1 == 0
E        +  where -1 = <Result TypeError('unaffected           0.0\n0ms - 999ms          0.0\n1000ms - 2296ms      1.0\n2297ms - 5276ms      ...124ms - 27855ms    0.0\n27856ms - 63999ms    0.0\n64000ms+             0.0\ndtype: float64 is not JSON serializable',)>.exit_code
```

### DIFF
--- a/mozetl/tab_spinner/generate_counts.py
+++ b/mozetl/tab_spinner/generate_counts.py
@@ -7,6 +7,7 @@ from moztelemetry.dataset import Dataset
 
 from .utils import get_short_and_long_spinners
 
+
 # numpy.float64 is not JSON serializable, so an encoder is provided for
 # serialization. In this particular case, the values are within a pandas series.
 # See: https://stackoverflow.com/a/47626762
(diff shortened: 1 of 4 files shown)


acmiyaguchi:
Testing this locally using this command:

```
python bin/mozetl-databricks.py \
    --git-path https://github.com/acmiyaguchi/python_mozetl.git \
    --git-branch tab-spinner \
    --python 3 \
    --token <TOKEN> \
    long_tab_spinners
```

Which should be visible here: https://dbc-caf9527b-e073.cloud.databricks.com/#job/1429/run/1

PR #320, opened 2019-03-21
Add content-based version detection.

mreid-moz:


mreid-moz:
@acmiyaguchi RFAL

mreid-moz:
Oops, already has r+, doesn't need another look :)

PR #326, opened 2019-04-15
Bug 1519233 - Add ad click counts to search datasets

wlach:
@harterrt could you tell me if this is on the right track? Basically I try to coerce the ad click counts into a similar format as the search_counts data in a seperate dataframe, then perform a union onto the main dataset. It *seems* to work, though this is pushing the limits of my pyspark knowledge/acumen.

If this looks ok, I'll parametrize this logic and also add a similar transformation for `scalar_parent_browser_search_with_ads`

### DIFF
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -30,7 +30,8 @@ from pyspark.sql.functions import (
     max,
     lit,
 )
-from pyspark.sql.types import StringType
+from pyspark.sql.types import LongType, ArrayType, StringType, StructType, StructField
+from pyspark.sql.utils import AnalysisException
 from pyspark.sql import SparkSession
 
 from mozetl.constants import SEARCH_SOURCE_WHITELIST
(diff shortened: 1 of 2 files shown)


mozetl/search/aggregates.py
```diff
@@ -171,17 +171,29 @@ def get_search_addon_version(active_addons):
     )
 
 
+def get_ad_click_count(ad_click_count):
+  if ad_click_count:
+    return [{'engine': e, 'source': "ad-click:", 'count': c} for e, c in ad_click_count.items()]
+  return []
```
harterrt: To confirm, users without ad-clicks will show up as `NULL`, correct? We impute search counts for 0 search users, but I don't think we can do that for ad-clicks since a large portion of clients don't have ad-click telemetry yet.

wlach: Yup.

wlach: Correct, I think.

wlach:
@harterrt I think this is ready for another look. I'll be out tomorrow + all next week, so no rush.

mozetl/search/aggregates.py
```diff
@@ -223,6 +305,8 @@ def _generate_when_expr(source_mappings):
                     ("in-content:organic:", "organic"),
                     ("sap:", "tagged-sap"),
                     ("follow-on:", "tagged-follow-on"),
+                    ("ad-click:", "ad-click"),
```
harterrt: Looks like this necessitates a change to the function docs on line 204.

mozetl/search/aggregates.py
```diff
@@ -171,28 +185,96 @@ def get_search_addon_version(active_addons):
...
+    def _drop_source_columns(base):
+        for source_col in [
+            "search_counts",
+            "scalar_parent_browser_search_ad_clicks",
+            "scalar_parent_browser_search_with_ads",
+        ]:
+            base = base.drop(source_col)
+        return base
+
+    def _get_ad_counts(scalar_name, col_name, udf_function):
```
harterrt: It feels like `_get_ad_counts` is doing something very similar to the work at lines 245. Is this generalizing the explode we do on `SEARCH_COUNTS` to apply to the ad-click probes? If so, can we refactor `explode_search_counts` so that the three explodes for `SEARCH_COUNTS`, ad-clicks, and searches-with-ads are all handled consistently?

wlach: Tried to do some refactoring into a common function. The code is rather nested and convoluted for my liking, but what can you do. :/

wlach:
@harterrt ready for another review I think

wlach:
Oops, I guess harter is away on PTO. I think I'm just going to go ahead and merge this, since he seemed ok with it other than the nits.

PR #328, opened 2019-04-23
Bug 1545799 - Add fields to search_clients_daily

relud:


relud:
cc @ophie200

mozetl/search/aggregates.py
```diff
@@ -50,57 +51,85 @@ def agg_first(col):
...
+                    ).alias("max_concurrent_tab_count_max")
+                ),
+                (
+                    sum("scalar_parent_browser_engagement_tab_open_event_count").alias(
+                        "tab_open_event_count_sum"
+                    )
+                ),
+                (sum(col("active_ticks") * 5 / 3600.0).alias("active_hours_sum")),
+                (
+                    collect_list("scalar_parent_browser_search_ad_clicks").alias(
```
relud: This is causing an error `expression 'search_ad_clicks_sum' cannot be used as a grouping expression because its data type array<map<string,int>> is not an orderable data type.`, @acmiyaguchi any ideas how to work around that? preferrably without rewriting this in spark >_>

acmiyaguchi: This is tricky. The pivot aggregation is causing the issues here:
https://github.com/mozilla/python_mozetl/blob/97cb6c243996c2d242a3b427878b19c10e18f177/mozetl/search/aggregates.py#L181-L186

It might be possible to run the aggregation function after the pivot. If not, I think the columns probably need to be exploded in a similar way to `explode_search_counts`.

harterrt:
Heads up, @wlach  is doing similar work in https://github.com/mozilla/python_mozetl/pull/326/

relud:
@harterrt thanks! closing this in favor of #326

PR #330, opened 2019-04-29
Bug 1547777 - Remove client count hll column from search_aggregates

wlach:
It didn't work correctly when aggregating across dimensions and was
unused. Best to just use search_clients_daily for this use case.

### DIFF
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -13,9 +13,7 @@ This allows for deeper analysis into user level search behavior.
 import click
 import logging
 import datetime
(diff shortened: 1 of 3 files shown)


wlach:
@harterrt review ping

wlach:
Hey @acmiyaguchi, is this something you could take a quick look at? See the bug for context. @harterrt already signed off on removing this, so we basically just need to make sure I'm not doing anything crazy.

wlach:
Thanks! I'll merge this now, can backfill later.

PR #331, opened 2019-05-30
Bug 1555764 - Add os, os_version columns to search datasets

wlach:


wlach:
I don't think it can hurt too much to merge this, so I'll do so.

PR #332, opened 2019-06-04
Normandy health check

glasserc:
This is somewhat WIP. This demonstrates the logic that I think I want, but I wasn't really clear on what the `main` function should look like, so that part is still up in the air. The overall idea is to set up monitoring tools like Pingdom to help ensure Normandy client is functioning correctly; to this end, Pingdom would monitor a thin service (unwritten yet) which would read data from S3 which had been ETLed by this job (run from Airflow).

This is my first analysis, use of Spark, etc. so please critique generously.

### DIFF
--- a/mozetl/health_check/normandy/job.py
+++ b/mozetl/health_check/normandy/job.py
@@ -1,0 +1,204 @@
+"""Normandy Health Check
+
+As part of our efforts to ensure that Normandy is continuing to
+function correctly, this analysis calculates some metrics based on
+uptake telemetry. This analysis is not itself responsible for
+reporting its results. Instead, it writes a summary to OUTPUT_FILENAME
+(a JSON file in S3). This file is monitored by a simple web service,
+which reports status as a 200/500 "heartbeat". This heartbeat is
+itself monitored using standard ops tools.
+
+"""
+
+import logging
+import arrow
+import boto3
+import click
+import json
+from pyspark.sql import SparkSession, functions as F, types as T
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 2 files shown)


mozetl/health_check/normandy/job.py
```diff
@@ -0,0 +1,204 @@
...
+        now, recipe_versions, recipe_health, action_health, runner_health, status_totals
+    )
+
+
+@click.command()
+@click.option("--period", default=6, help="length of the retention period in hours")
+@click.option(
+    "--slack", default=6, help="number of hours to account for submission latency"
+)
+def main(period, slack):
```
wlach: I would consider putting this function outside of click, so it's easier to run the code from e.g. a databricks notebook. You can see an example of this in search aggregates:

https://github.com/mozilla/python_mozetl/blob/ad257f64c5de876b615e7ba56608d0c91d7d2fe6/mozetl/search/aggregates.py#L390

mozetl/health_check/normandy/job.py
```diff
@@ -0,0 +1,204 @@
...
+    events_start = now.shift(hours=-slack)
+    longitudinal = spark.sql(
+        """
+        SELECT client_id, submission_date, scalar_parent_normandy_recipe_freshness
+        FROM longitudinal
+        """
+    )
+    longitudinal_start = events_start.shift(days=-1)
+    result = extract(events, events_start, longitudinal, longitudinal_start)
+    s3.put_object(Body=json.dump(result), Bucket=OUTPUT_BUCKET, Key=OUTPUT_KEY)
```
wlach: I would make these parameters of the function when extracting it out, that way you can easily write test output to a test bucket / directory.

glasserc:
In Whistler, I was encouraged to instead add this to https://github.com/mozilla/bigquery-etl/, so I'm going to close this PR.

PR #333, opened 2019-07-08
Fix overcounting error when exploding ad fields

benmiroglio:
I found an error when looking at the ad probes using search clients daily versus main_summary. There is an over-counting error when exploding those fields.

The error/fix can be more easily explored in this minimal reproduction:

https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/141261/command/141311

@harterrt r?

### DIFF
--- a/mozetl/search/aggregates.py
+++ b/mozetl/search/aggregates.py
@@ -221,22 +221,21 @@ def explode_search_counts(main_summary):
             derived = derived.drop(source_col)
         return derived
 
-    def _select_counts(
-        main_summary, col_name, count_udf=None, get_single_search_count=True
-    ):
-        if get_single_search_count:
+    def _select_counts(main_summary, col_name, count_udf=None):
+        ad_fields = ("ad_click", "search_with_ads")
+        if col_name in ad_fields:
+            # special case for ad click probes
+            derived = main_summary
+        else:
             derived = main_summary.withColumn(
                 "single_search_count", explode(col("search_counts"))
             ).filter("single_search_count.count < %s" % MAX_CLIENT_SEARCH_COUNT)
-        else:
-            # special case for ad-click and search-with-ads probes
-            derived = main_summary
         if count_udf is not None:
             derived = derived.withColumn(col_name, explode(count_udf))
         derived = _drop_source_columns(
             derived.select(["*"] + _get_search_fields(col_name)).drop(col_name)
         )
-        if col_name is not "single_search_count":
+        if col_name is not "single_search_count" or col_name in ad_fields:
             derived = derived.drop("single_search_count")
         return derived
 
@@ -253,12 +252,7 @@ def explode_search_counts(main_summary):
                 )
             ),
         )
-        return _select_counts(
-            main_summary,
-            col_name,
-            count_udf(scalar_name),
-            get_single_search_count=False,
-        )
+        return _select_counts(main_summary, col_name, count_udf(scalar_name))
 
     exploded_search_counts = _select_counts(main_summary, "single_search_count")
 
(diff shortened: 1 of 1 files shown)


benmiroglio:
also tagging @wlach for review since you've worked on this dataset recently.

mozetl/search/aggregates.py
```diff
@@ -221,10 +221,16 @@ def _drop_source_columns(base):
             derived = derived.drop(source_col)
         return derived
 
-    def _select_counts(main_summary, col_name, count_udf=None):
-        derived = main_summary.withColumn(
-            "single_search_count", explode(col("search_counts"))
-        ).filter("single_search_count.count < %s" % MAX_CLIENT_SEARCH_COUNT)
+    def _select_counts(
+        main_summary, col_name, count_udf=None, get_single_search_count=True
```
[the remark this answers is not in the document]
benmiroglio: Certainly not, it was just the minimal amount of code needed to fix the issue. It probably makes sense to break the aggregations out for clarity.

benmiroglio:
@wlach I removed the parameter and used the column name to apply the same logic. Breaking this out into two different functions seemed like overkill after prototyping. 

Regarding adding tests, I took a look at the tests and it became clear it will require a good amount of cognitive load for me to figure out how to properly account for this edge case. I can get around to it eventually but I'll defer to you on whether we should block this PR until the test properly account for this.

mozetl/search/aggregates.py
```diff
@@ -222,15 +222,20 @@ def _drop_source_columns(base):
...
-            "single_search_count", explode(col("search_counts"))
-        ).filter("single_search_count.count < %s" % MAX_CLIENT_SEARCH_COUNT)
+        ad_fields = ("ad_click", "search_with_ads")
+        if col_name in ad_fields:
+            # special case for ad click probes
+            derived = main_summary
+        else:
+            derived = main_summary.withColumn(
+                "single_search_count", explode(col("search_counts"))
+            ).filter("single_search_count.count < %s" % MAX_CLIENT_SEARCH_COUNT)
```
wlach: I don't think we really need to specify the ad click fields. Really it's the single search count case which is "special". I think I would prefer something like this:

```py
if col_name == "single_search_count":
    derived = main_summary.withColumn(
                "single_search_count", explode(col("search_counts"))
            ).filter("single_search_count.count < %s" % MAX_CLIENT_SEARCH_COUNT)
else:
    derived = main_summary
```

mozetl/search/aggregates.py
```diff
@@ -222,15 +222,20 @@ def _drop_source_columns(base):
...
         if count_udf is not None:
             derived = derived.withColumn(col_name, explode(count_udf))
         derived = _drop_source_columns(
             derived.select(["*"] + _get_search_fields(col_name)).drop(col_name)
         )
-        if col_name is not "single_search_count":
+        if col_name is not "single_search_count" or col_name in ad_fields:
```
wlach: I don't think this extra logic is necessary. If the col_name is one of the ad fields, by definition it is not "single_search_count".

benmiroglio: True, this will just drop a phantom column in the case of the ad_clicks, but I was being (perhaps too) explicit.

mozetl/search/aggregates.py
benmiroglio: You're right, this is cleaner!

mozetl/search/aggregates.py
benmiroglio: Actually, I think this if-statement can be omitted altogether with the added logic at the top of the function.

wlach:
This looks good! Thank you Ben. We should probably backfill the data into the past to make sure we're using valid data here, I'll file a bug about that tomorrow. For now, I'll merge this.

PR #334, opened 2019-07-17
TAAR whitelist uses recommended addons only now.

crankycoder:
The AMO whitelist URL is now requested with a 'recex=True' option added
to the end of the URL to restrict addons to only the recommended list
for TAAR recommendd addons now.

cc @muffinresearch

### DIFF
--- a/mozetl/taar/taar_update_whitelist.py
+++ b/mozetl/taar/taar_update_whitelist.py
@@ -25,8 +25,13 @@ class GUIDError(BaseException):
     pass
 
 
-def load_raw_json(url):
-    r = requests.get(url)
+def load_amo_editorial(url, only_recommended=True):
+    param_dict = {}
+
+    if only_recommended:
+        param_dict["recommended"] = "true"
+
+    r = requests.get(url, params=param_dict)
     if 200 == r.status_code:
         # process stuff here
         json_data = json.loads(r.text)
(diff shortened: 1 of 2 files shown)


crankycoder:
This patch also had to update the CircleCI yaml configuration to explicitly pin the use of Debian 9 (stretch) instead of the default Debian build which is now Debian 10 (buster). 

Debian Buster does not officially support OpenJDK 8 anymore - so everything broke.

.circleci/config.yml
```diff
@@ -70,30 +70,33 @@ jobs:
...
+      - image: python:3.6-stretch
     working_directory: ~/python_mozetl
     steps:
       - checkout
       - run: *early_return_for_skip_tests
       - restore_cache: *restore_cache_settings
       - run: *install_dependencies
+      - run:
+          name: install flake8 and black
+          command: pip install flake8 black
```
acmiyaguchi: This step should be unnecessary because they;re listed as dependencies in `tox.ini`

mozetl/taar/taar_update_whitelist.py
```diff
@@ -74,11 +79,11 @@ def load_etl(transformed_data, date, prefix, bucket):
 @click.command()
 @click.option("--date", required=True)
 @click.option("--url", default=EDITORIAL_URI)
+@click.option("--only_recommended", default=True)
```
acmiyaguchi: nit: kebab case is preferred (hyphens), since most cli tools tend to use it.

.circleci/config.yml
crankycoder: Odd - I must've had an unrelated problem causing flake8 to not be found.  I've removed the step now.

crankycoder:
thanks @acmiyaguchi !  let's wait until the addons-server [patch](https://github.com/mozilla/addons-server/issues/11851) before merging this in.

eviljeff:
the addons-server patch has landed on addons-dev.allizom.org now (it'll be on addons.allizom.org on Tuesday; production addons.mozilla.org on July 25th)

crankycoder:
@acmiyaguchi this should be ok to land now.  flags at 

https://github.com/mozilla/python_mozetl/blob/cc09ba189ade26bc675de75d29f3c0c0ef01fe74/mozetl/taar/taar_update_whitelist.py#L32

and at

https://github.com/eviljeff/olympia/blob/8483d30f7cfdef929b9980ebeb94db7453d07ac9/src/olympia/discovery/views.py#L92

muffinresearch:
The AMO patch should be in production now.

PR #335, opened 2019-07-18
Pins the docker images to use debian 9 (stretch)

crankycoder:
* CircleCI just upgraded to debian 10 (buster) which does not not support
OpenJDK 8.


crankycoder:
I've cherry-picked the circle fixes out of #334 so that we can get CI working again.

crankycoder:
@acmiyaguchi r?

PR #336, opened 2019-08-01
updated taar whitelist to allow for short recommended whitelists

crankycoder:
The new recommended addon list from AMO is quite short (<90 addons), this relaxes that constraint.

### DIFF
--- a/mozetl/taar/taar_update_whitelist.py
+++ b/mozetl/taar/taar_update_whitelist.py
@@ -82,8 +82,9 @@ def load_etl(transformed_data, date, prefix, bucket):
 @click.option("--only-recommended", default=True)
 @click.option("--bucket", default="telemetry-parquet")
 @click.option("--prefix", default="taar/locale/")
-@click.option("--validate_guid", default=False)
-def main(date, url, only_recommended, bucket, prefix, validate_guid):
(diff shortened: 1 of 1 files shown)


crankycoder:
The short whitelist broke an old safety guard which checked that TAAR always has at least 100 addons in its recommendation pool.

This patch relaxes that constraint by default.

We missed this during last weeks patch as we did not have a live API in AMO to test against and I wasn't expecting the list to be so short.

I've tested this patch with mozetl-submit.sh

```
(aws)  vng@Victor-MBPrTouch  ~/dev/python_mozetl  aws s3 ls s3://telemetry-parquet/taar/locale/only_guids_top_20020190731.json
2019-08-01 13:17:47       3186 only_guids_top_20020190731.json
```

crankycoder:
@acmiyaguchi the addons editorial team noticed recommendations for addons in about:addons that should not be happening.

This was actually two separate problems:

1) covered by this patch, we did not deal with whitelists < 100 elements long.  During the initial SHIELD experiments, a whitelist of such a short length indicated that something had gone really wrong.

2) We've had multiple variations of the whitelist generated by different processes between taar and taar-lite.  The write path was set incorrectly which is addressed in: https://github.com/mozilla/python_mozetl/pull/337

PR #337, opened 2019-08-01
fixes the s3 write path for the recommended whitelist

crankycoder:
The default write path for our whitelist was set incorrectly.


crankycoder:
I've updated the default path and rerun using mozetl-submit.sh to verify the output results for today's run.

```
(aws)  vng@Victor-MBPrTouch  ~/dev/python_mozetl/mozetl/taar  aws s3 ls s3://telemetry-parquet/telemetry-ml/addon_recommender/only
2019-08-01 15:03:33       3186 only_guids_top_200.json
2019-01-08 13:39:40       5766 only_guids_top_200.json.tmp
2019-08-01 15:03:33       3186 only_guids_top_20020190801.json```

PR #338, opened 2019-08-13
Bug 1571565 - Decommission retention_v1 dataset

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1571565)


acmiyaguchi:
Nope, churn is going to go away to reduce the Spark workload in GCP. Submission date is the preferred anchored for this kind of analysis now (which reduces the overhead of the "window" period), so keeping it would be potentially confusing.

PR #343, opened 2019-10-21
Fix undercounting of tab spinners

wlach:
We would consider something "unaffected" for a short duration if we had no "short" tab spinner histogram OR "long" tab spinner histograms. It should
actually be "no short AND no long".


wlach:
Noticed this while working on https://bugzilla.mozilla.org/show_bug.cgi?id=1589118. I think this case is rather unlikely, if not impossible, but it is probably best to fix it regardless. /cc @mikeconley

wlach:
This query seems to confirm this situation is a hypothetical: https://sql.telemetry.mozilla.org/queries/65494/source

PR #351, opened 2020-11-26
Port graphics trends from databricks to dataproc

BenWu:
Migration of graphics-telemetry-dashboard-big-query and graphics-telemetry-trends-big-query jobs from databricks.  Intention is to use this repo as storage for the python driver code for a dataproc pyspark job.  Associated airflow pr https://github.com/mozilla/telemetry-airflow/pull/1191

Code is mostly unchanged; this is using boto3 for s3 access and adds cli arguments

The notebooks:
https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/239724
https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/240233

circleci deploy keys for this repo were deleted so ci doesn't work but I don't think it's needed.

### DIFF
--- a/mozetl/graphics/create_cluster.sh
+++ b/mozetl/graphics/create_cluster.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+if [[ -z $AWS_ACCESS_KEY_ID ]]; then
+    echo '$AWS_ACCESS_KEY_ID not defined'
+    exit 1
+fi
+if [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
+    echo '$AWS_SECRET_ACCESS_KEY not defined'
+    exit 1
+fi
+
+gcloud dataproc clusters create graphics-spark-test \
+    --image-version=1.5 \
+    --region=us-central1 \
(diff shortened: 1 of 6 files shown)


mozetl/graphics/create_cluster.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
akkomar, line 1: Are these scripts for testing? It's not clear, some comment about this would be nice.

akkomar (changes requested):
Notebook code looks good, let's not dive deep into that :)

One nit about scripts - I guess they're for testing purposes?

> circleci deploy keys for this repo were deleted so ci doesn't work but I don't think it's needed.

We should fix that before merging (linting can fail here).

mozetl/graphics/create_cluster.sh
BenWu: Scripts are for testing, I'll add a comment.

BenWu:
I formatted code and updated dependencies in setup.py.  Still some failing ci for other reasons. I'm at eod and can't look closer but I'll revisit tomorrow.

akkomar (approved):
It's not immediately clear to me what's wrong with the tests.
I'm giving r+ because I think we're under time pressure here.
Later on we should figure out what's wrong with the build.

BenWu:
I created issue https://github.com/mozilla/python_mozetl/issues/352 listing the CI errors.  I'll take another look later

Issue #352, opened 2020-12-11
Fix CI


BenWu:
There are various new issues with circleci since this repo was last used.  

linter times out:
https://app.circleci.com/pipelines/github/mozilla/python_mozetl/13/workflows/cc6842ee-8f4e-426b-856a-d96c82904041/jobs/1244

coverage failure:
https://app.circleci.com/pipelines/github/mozilla/python_mozetl/13/workflows/cc6842ee-8f4e-426b-856a-d96c82904041/jobs/1245/parallel-runs/0?filterBy=FAILED

tests using s3 (boto3 and moto versions were bumped in #351 to fix dependency resolution errors, may be related):
https://app.circleci.com/pipelines/github/mozilla/python_mozetl/13/workflows/cc6842ee-8f4e-426b-856a-d96c82904041/jobs/1245/parallel-runs/2?filterBy=FAILED

tests for py3.5 time out:
https://app.circleci.com/pipelines/github/mozilla/python_mozetl/13/workflows/cc6842ee-8f4e-426b-856a-d96c82904041/jobs/1243

PR #354, opened 2020-12-16
Update dependencies to fix ci

BenWu:
(fixes https://github.com/mozilla/python_mozetl/issues/352)

Mostly just updating dependency versions and python version.  Using python 3.7 because that's the last version that has an image built on debian 9 which allows `apt install openjdk-8-jre-headless`.  There are other ways to get jdk on later versions but I don't think it's worth the effort and complexity since 3.7 will be relevant for another couple of years.

PR also has black/flake8 fixes

Removed tests for python2 because I don't see any reason to keep them


akkomar (approved):
LGTM, than you!

PR #356, opened 2020-12-17
Port bhr collection from databricks

BenWu:
Source notebook: https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/211711/command/216952


mozetl/bhr_collection/bhr_collection.py
```diff
@@ -0,0 +1,1359 @@
...
+from pyspark import SparkContext
+from pyspark.sql import SparkSession
+from pyspark.sql.functions import array, collect_list
+from pyspark.sql.types import Row
+
+UNSYMBOLICATED = "<unsymbolicated>"
+SYMBOL_TRUNCATE_LENGTH = 200
+
+sc = SparkContext.getOrCreate()
+spark = SparkSession.builder.appName("bhr-collection").getOrCreate()
```
acmiyaguchi, line 32:
```suggestion
spark = SparkSession.builder.appName("bhr-collection").getOrCreate()
sc = spark.sparkContext
```

mozetl/bhr_collection/bhr_collection.py
```diff
@@ -0,0 +1,1359 @@
...
+
+    submission_start_str = date - timedelta(days=5)
+    submission_end_str = end_date + timedelta(days=5)
+
+    date_str = date.strftime("%Y%m%d")
+    end_date_str = end_date.strftime("%Y%m%d")
+
+    pings_df = (
+        spark.read.format("bigquery")
+        .option("table", "moz-fx-data-shared-prod.telemetry_stable.bhr_v4")
```
acmiyaguchi, line 516: Is this using the new structured format as introduced in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/636 on 2020-12-07? From what I'm reading, it looks like it does due to the way that the stacks are being parsed, but I just want to make sure.

mozetl/bhr_collection/bhr_collection.py
```diff
@@ -0,0 +1,1359 @@
...
+        print("writing file %s" % filepath)
+        write_file(filepath, profile, final_config)
+        filepath = "%s_current" % final_config["hang_profile_out_filename"]
+        print("writing file %s" % filepath)
+        write_file(filepath, profile, final_config)
+        gc.collect()
+        print_progress(job_start, iterations, x, iteration_start, date_str)
+
+
+def etl_job_incremental_finalize(_, __, config=None):
```
acmiyaguchi, line 1287: This is a very strange definition signature. Also this function seems to be unused?

mozetl/bhr_collection/bhr_collection.py
```diff
@@ -0,0 +1,1359 @@
...
+    default=0.5,
+    help="Proportion of pings to use (1.0 is 100%)",
+)
+def start_job(sample_size):
+    print(f"Using sample size {sample_size}")
+    etl_job_daily(
+        sc,
+        spark,
+        {
+            "start_date": datetime.today() - timedelta(days=3),
```
acmiyaguchi, line 1347: This command should take the start_date parameter as an argument so the Airflow dag can take over what date ranges are run.

acmiyaguchi (approved):
It would useful to include a README at the bhr_collection level on how to invoke the job manually.

r+, but I strongly suggest making the `start_date` a parameter for the airflow job.

mozetl/bhr_collection/bhr_collection.py
BenWu: Yes this is what broke the job last week and needed to get fixed before moving it over.

mozetl/bhr_collection/bhr_collection.py
BenWu: Yes that looks like some weird artifact of working in a notebook. This and a few other etl_job functions are unused so I'll take them out.  My code inspector missed these for some reason.

PR #358, opened 2021-02-17
Change graphics jobs to use mozdata for temp tables

BenWu:
Changes to use mozdata for analysis tables.  I'll make a pr to them shim code as well for completeness


jklukas (approved):
Sorry for missing this earlier in the week. Looks good 👍

PR #361, opened 2021-08-17
Fix error on missing media decoders in graphics job

BenWu:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1694448#c4

If this is all it takes to fix then probably should have done this sooner


BenWu:
added lines, diff coverage is 0% => coverage increases by 3.53%
ty codecov

jklukas:
Ship it!

PR #362, opened 2021-12-20
bug 1746874: fix case where version is None

willkg:
Socorro switched stackwalkers last week and the new one emits a null
instead of an empty string for the module version. This fixes
`is_old_firefox_module` to work with null (which is None).


jklukas (approved):

willkg:
Oops--I merged that with a merge commit. I didn't see other options, but the project settings suggest I should have. Bleh.

PR #367, opened 2022-03-19
Generate correlations for the top 1000 signatures instead of top 300

marco-c:


willkg:
What's the impetus behind this change? Any idea how it will affect runtime?

marco-c:
It's just to have correlations available for more crashes.
The runtime shouldn't change much because most of the time should be spent on getting statistics about *all* crashes, which is done anyway no matter how many signatures the script analyzes. I didn't actually test though, so I might be wrong (I guess we can keep an eye on it after, and if, we merge this).

willkg:
If there are no downsides to this change and the goal is to have correlations available for more crashes, why raise it to 1,000? Why not raise it to 10,000?

marco-c:
Two reasons:
1) It's unlikely that anyone is going to look at such small volume crashes;
2) The script is ignoring signatures which have very few crashes because the correlations would be spurious (simply due to the low volume you'd see correlations that do not actually exist).

willkg:
The top_signatures_correlations code runs monday, wednesday, and friday and it takes about 1h20m to run.

Looking at the output, seems like the release channel is the only one that might be affected by this change--beta, nightly, and esr won't be because they are already picking up signatures for which there aren't enough crash reports.

This seems fine. I'll check in on it tomorrow and see what happened.

willkg (approved):

marco-c:
All right, it's way worse than I thought. Let's revert to 300.

willkg:
I'll do that today. Thank you!

PR #370, opened 2022-05-04
Allow empty symbols in breakpad files for BHR

squarewave:
The bhr_collection job is currently failing because we are now
seeing a FUNC line in a breakpad file which omits a symbol. I'm
not entirely sure why we're seeing this, but I don't see any
problem with just letting it through annotated as an empty symbol.

### DIFF
--- a/mozetl/bhr_collection/bhr_collection.py
+++ b/mozetl/bhr_collection/bhr_collection.py
@@ -919,10 +919,13 @@ def make_sym_map(data):
             if fields[1] == "m":
                 m_offset = 1
                 fields = line.split(" ", 5)
-            if len(fields) < 5 + m_offset:
+            if len(fields) == 4 + m_offset:
+                symbol = "(no symbol)"
+            elif len(fields) < 4 + m_offset:
                 raise ValueError("Failed to parse address - line: {}".format(line))
+            else
+                symbol = fields[4 + m_offset]
             address = int(fields[1 + m_offset], 16)
-            symbol = fields[4 + m_offset]
             func_symbols[address] = symbol[:SYMBOL_TRUNCATE_LENGTH]
     # Prioritize PUBLIC symbols over FUNC ones
     sym_map = func_symbols


willkg:
I didn't know about this etl thing. We have a Symbolication API that'll handle fetching sym files from the symbol server, parsing them, and doing lookups. The Mozilla Symbolication Server is what we're going to use for symbolicating stacks in crash pings. Further, it's actively maintained and aligned with symbol file generation, storage, and other tooling around symbols.

I'm pretty sure it would help here and it'd both reduce the maintenance burden plus make it easier to stay aligned and supported by the rest of the reporting machinery.

https://symbolication.services.mozilla.com/

Let me know if you want help with this.

willkg:
Also, I'm working on moving the Mozilla Symbols Server from AWS to GCP this year which will break this script since it's got the AWS S3 bucket hard-coded:

https://github.com/mozilla/python_mozetl/blob/5435e3066a71b7b402da771ada4ab586edbc330e/mozetl/bhr_collection/bhr_collection.py#L1142-L1143

If you want to keep doing your own symbolication, you should follow the work in this bug so this script doesn't stop working when we switch to GCP:

https://bugzilla.mozilla.org/show_bug.cgi?id=1687804

squarewave:
> I didn't know about this etl thing. We have a Symbolication API that'll handle fetching sym files from the symbol server, parsing them, and doing lookups. The Mozilla Symbolication Server is what we're going to use for symbolicating stacks in crash pings. Further, it's actively maintained and aligned with symbol file generation, storage, and other tooling around symbols.
> 
> I'm pretty sure it would help here and it'd both reduce the maintenance burden plus make it easier to stay aligned and supported by the rest of the reporting machinery.
> 
> https://symbolication.services.mozilla.com/
> 
> Let me know if you want help with this.

Yeah - this job was originally using the old symbolication API, but the volume was way too high for it. I tried once in the intervening years to switch to the newer API, but as I recall I ran into similar issues. The volume here is really high - substantially higher than the present volume from either crash reporting or the profiler UI. I'm sure it's workable, but I suspect it would perform worse than doing it in-job, even though it's just in python. It would be nice though if it would allow us to get more representative stacks, but purely to reduce maintenance burden I don't know if the up-front investment in exploring pointing it at the API is worth it. I think this is the first time I've really touched this code in a few years?

willkg:
Ok, that makes sense. The symbolication server is new--I only stood it up at the end of last year.

I would highly suggest switching to using symbolic for parsing sym files and doing lookups. That's what the symbolication server is doing. If you can keep the symcache files around, they load much faster than re-parsing sym files.

https://pypi.org/project/symbolic/

jklukas (approved):
This seems like a reasonable update to get this unstuck although please do consider @willkg's suggestions for further improvement.

mozetl/bhr_collection/bhr_collection.py
```diff
@@ -919,10 +919,13 @@ def make_sym_map(data):
             if fields[1] == "m":
                 m_offset = 1
                 fields = line.split(" ", 5)
-            if len(fields) < 5 + m_offset:
+            if len(fields) == 4 + m_offset:
+                symbol = "(no symbol)"
+            elif len(fields) < 4 + m_offset:
                 raise ValueError("Failed to parse address - line: {}".format(line))
+            else
```
jklukas, line 926: Turns out this was missing a `:` here.

PR #372, opened 2022-06-15
bug 1774415: reduce top signatures to 250

willkg:
The top_signatures_correlations task is getting killed off for using too
much memory. This drops the top signatures we're looking at from 300 to
250 as a temporary attempt to get the job succeeding again.

I also did some "magic number" reduction so it reads better.


jklukas (approved):
Looks reasonable

willkg:
I'll keep an eye on this and see what happens.

Thank you!

PR #376, opened 2022-11-02
add README for symbolication scripts

willkg:
This adds some rough documentation for the symbolication scripts covering what they do, why, and how it fits in with the world.


willkg:
This repository has a Sphinx docs project where module-level documentation is extracted, but that's all in `source` and that's not available from the index page, so I don't think any of that works. I didn't update module documentation.

However, regardless of whether modules are documented well, it's good to have README files so that people looking at the files have something they can rely on to get a lay of the land.

mreid-moz (approved):
👍 This adds useful info in a place that's easy to find.

willkg:
Thank you!

Issue #373, opened 2022-06-15
tests fail: TypeError: Descriptors cannot not be created directly.


willkg:
Making any updates in this repo kicks up a bunch of CI test failures:

```
collection failure
tests/test_utils.py:8: in <module>
    from mozetl import utils
mozetl/__init__.py:1: in <module>
    from .main import *  # noqa
mozetl/main.py:4: in <module>
    from moztelemetry import get_pings_properties
.tox/py37/lib/python3.7/site-packages/moztelemetry/__init__.py:2: in <module>
    from .spark import *
.tox/py37/lib/python3.7/site-packages/moztelemetry/spark.py:16: in <module>
    from .dataset import Dataset
.tox/py37/lib/python3.7/site-packages/moztelemetry/dataset.py:19: in <module>
    from .heka import message_parser
.tox/py37/lib/python3.7/site-packages/moztelemetry/heka/message_parser.py:16: in <module>
    from .message_pb2 import Message, Header
.tox/py37/lib/python3.7/site-packages/moztelemetry/heka/message_pb2.py:36: in <module>
    type=None),
.tox/py37/lib/python3.7/site-packages/google/protobuf/descriptor.py:755: in __new__
    _message.Message._CheckCalledFromGeneratedFile()
E   TypeError: Descriptors cannot not be created directly.
E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
E   If you cannot immediately regenerate your protos, some other possible workarounds are:
E    1. Downgrade the protobuf package to 3.20.x or lower.
E    2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
E   
E   More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates

willkg:
It looks like the moztelemetry library was archived 3 years ago:

https://github.com/mozilla/python_moztelemetry

Is this even fixable without unarchiving that library and updating it?

PR #379, opened 2023-01-16
Pin flake8 and protobuf to fix lint and py37 CI jobs

ANich:
Dependency update for CI fix.

`flake8` for the `lint` job - 4.0.0 because it pins `importlib` due to a breaking change which caused the `AttributeError: 'EntryPoints' object has no attribute 'get'` problem. 

`protobuf` for the tests - 3.20.3 because that's what's suggested in the warnings:
```
E   TypeError: Descriptors cannot not be created directly.
E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
E   If you cannot immediately regenerate your protos, some other possible workarounds are:
E    1. Downgrade the protobuf package to 3.20.x or lower.
E    2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
E   
E   More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

Closes https://github.com/mozilla/python_mozetl/issues/373

Probably worth keeping an eye on airflow as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ test_deps = [
     'moto==1.3.16',
     'mock==2.0.0',
     'pytest==3.10.1',
-    'flake8==3.8.4'
+    'flake8==4.0.0'
 ]
 
 extras = {
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
         'click_datetime==0.2',
         'numpy==1.19.4',
         'pandas==1.1.4',
+        'protobuf==3.20.3',
         'pyspark==2.3.2',
         'python-moztelemetry @ git+http://github.com/mozilla/python_moztelemetry.git@v0.10.2#egg=python-moztelemetry',
         'requests-toolbelt==0.9.1',
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ test_deps = [
     'moto==1.3.16',
     'mock==2.0.0',
     'pytest==3.10.1',
+    # NOTE: this is pinned to 4.0.0 so as to pin importlib to a version that works.
     'flake8==4.0.0'
 ]
 
@@ -34,6 +35,7 @@ setup(
         'click_datetime==0.2',
         'numpy==1.19.4',
         'pandas==1.1.4',
+        # NOTE: this is pinned to 3.20.3 because protos changed and we can't regenerate them.
         'protobuf==3.20.3',
         'pyspark==2.3.2',
         'python-moztelemetry @ git+http://github.com/mozilla/python_moztelemetry.git@v0.10.2#egg=python-moztelemetry',


willkg:
For flake8, they're saying that you should run flake8 in isolation and then you don't have to deal with the dependency conflicts with other applications:

https://github.com/PyCQA/flake8/issues/1522#issuecomment-1014558247

While we could pin it to an older version, I think it's probably never going to be conflict free for reasons and we're better off running flake8 in an isolated environment.

I don't know what the deal with protobuf is. What protobuf things are out of date?

ANich:
> While we could pin it to an older version, I think it's probably never going to be conflict free for reasons and we're better off running flake8 in an isolated environment.

Hmm don't think I have strong opinions about this, we could use a pre-commit hook which is one of the suggestions. We do this in bigquery-etl [here](https://github.com/mozilla/bigquery-etl/blob/fffa763f88e44cb6f009b77cf629de6a07e2f326/.pre-commit-config.yaml#L12). But this wouldn't fix it in CI so if we want to keep having a flake8 check there, we'll have to get a version there somehow (bigquery-etl pins a version [here](https://github.com/mozilla/bigquery-etl/blob/fffa763f88e44cb6f009b77cf629de6a07e2f326/requirements.txt#L273) for CI use). 

> I don't know what the deal with protobuf is. What protobuf things are out of date?

Haven't dug too far into this honestly. The stack trace suggests something upstream of pyspark but trying a newer version of pyspark didn't help so I just went with the first workaround suggested: "1. Downgrade the protobuf package to 3.20.x or lower."

setup.py
```diff
@@ -8,7 +8,7 @@
     'moto==1.3.16',
     'mock==2.0.0',
     'pytest==3.10.1',
-    'flake8==3.8.4'
+    'flake8==4.0.0'
```
willkg, line 11: We should add a comment here as to why this is pinned at this version so future us doesn't try to update it.

willkg, line 11: We need to add a comment here so future us knows why this is pinned to this specific version.

```suggestion
    # NOTE: this is pinned to 4.0.0 so as to pin importlib to a version that works.
    'flake8==4.0.0'
```

willkg:
I use tox which uses isolated environments and works in CI and in local development.

However, that's a big lift and too big for this PR. I think python_mozetl is deprecated, so I'm not sure it's worth spending a lot of time and energy fixing technical debt issues like this one. Maybe write up an issue so it's tracked somewhere, add the two comments I suggested, and then I'm r+ on this change.

willkg (approved):
Looks super! Thank you for adding the comments! Future us will be so much happier.

PR #390, opened 2023-10-03
DENG-1023 crash_symbolication stores results on gcs instead of s3 (dev bucket for testing)

edugfilho:
Along with https://github.com/mozilla/telemetry-airflow/pull/1829

This also removes the dependency on https://github.com/marco-c/crashcorrelations for storing result files in buckets.

Once I can confirm the files are written to the dev bucket I'll open another PR with the prod bucket name


willkg (approved):
I don't see any issues with this. Looks good!

PR #391, opened 2023-10-04
DENG-1023 crash_symbolication stores results on gcs instead of s3 (prod)

edugfilho:
The job worked in dev, so I'm updating the bucket with the prod one.

Note: last execution sent results to dev bucket, which I manually copied over to prod and tested by downloading some files from https://analysis-output.telemetry.mozilla.org


willkg (approved):
I don't know the names of the buckets. Assuming the changed name is correct, this looks good to me.

edugfilho:
> I don't know the names of the buckets. Assuming the changed name is correct, this looks good to me.

They're listed in this instruction doc: https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/285343855/Updating+a+DAG+to+write+to+GCS+instead+of+S3

PR #392, opened 2023-10-05
Fix the bhr script to handle the extra arguements

alekhyamoz:
The previous PR - https://github.com/mozilla/python_mozetl/pull/387 did not handle the `extra_args = {"ContentType": "application/json", "ContentEncoding": "gzip"}`. due to which the process was unable to read the files.


edugfilho (approved):
Although I don't think we need the `use_gcs` flag because the `s3` bucket will be decommissioned this lgtm. Thank you!

PR #399, opened 2024-01-08
 Add date in YYYY-mm-dd to end of email subject line

willkg:
This adds the date in YYYY-mm-dd to the end of the email subject line so as to fix threading in the google groups interface.

This mirrors the change here: https://github.com/marco-c/missing_symbols/commit/2c3a1b4cc83e0012c5c18e0379f144d8533d9419

### DIFF
--- a/mozetl/symbolication/modules_with_missing_symbols.py
+++ b/mozetl/symbolication/modules_with_missing_symbols.py
@@ -167,7 +167,10 @@ top_missing_with_avail_info = [
     for name, (version, debug_id, debug_file), count in top_missing
 ]
 
-subject = "Weekly report of modules with missing symbols in crash reports: %s" % datetime.today().strftime("%Y-%m-%d")
+subject = (
+    "Weekly report of modules with missing symbols in crash reports: %s"
+    % datetime.today().strftime("%Y-%m-%d")
+)
 
 body = """
 <table style="border-collapse:collapse;">
--- a/mozetl/symbolication/modules_with_missing_symbols.py
+++ b/mozetl/symbolication/modules_with_missing_symbols.py
@@ -167,7 +167,7 @@ top_missing_with_avail_info = [
     for name, (version, debug_id, debug_file), count in top_missing
 ]
 
-subject = "Weekly report of modules with missing symbols in crash reports"
+subject = "Weekly report of modules with missing symbols in crash reports: %s" % datetime.today().strftime("%Y-%m-%d")
 
 body = """
 <table style="border-collapse:collapse;">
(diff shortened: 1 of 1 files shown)


mozetl/symbolication/modules_with_missing_symbols.py
```diff
@@ -167,7 +167,10 @@ def are_symbols_available(debug_file, debug_id):
     for name, (version, debug_id, debug_file), count in top_missing
 ]
 
-subject = "Weekly report of modules with missing symbols in crash reports"
+subject = (
+    "Weekly report of modules with missing symbols in crash reports: %s"
+    % datetime.today().strftime("%Y-%m-%d")
```
ANich, line 172: Should this use the `--date` parameter defined above?

ANich, line 172: nit: 
```suggestion
subject = (
    f"Weekly report of modules with missing symbols in crash reports: "
    f"{datetime.today().strftime('%Y-%m-%d')}"
)
```

ANich (approved):
LGTM, left one question about the date value

mozetl/symbolication/modules_with_missing_symbols.py
willkg: The first f-string doesn't have anything to expand, so it won't pass the linter. I tend not to put function calls in f-strings because of the way f-strings get parsed by the Python parser--it's kind of wild though I think they improved that in 3.12 or 3.13.

I'll adjust it so it's in an intermediary string after you weigh in on whether we should use the `--date` arg value.

mozetl/symbolication/modules_with_missing_symbols.py
```diff
@@ -167,7 +167,10 @@ def are_symbols_available(debug_file, debug_id):
     for name, (version, debug_id, debug_file), count in top_missing
 ]
 
-subject = "Weekly report of modules with missing symbols in crash reports"
+today_date = datetime.today().strftime("%Y-%m-%d")
```
willkg, line 170: This uses `datetime.today()` because that's what the script is querying with. The script takes a `date` param, but that's only used to determine whether the script should run and isn't used for the queries.

mozetl/symbolication/modules_with_missing_symbols.py
willkg: I added a note about how we should use `datetime.today()` as the script is currently built.

willkg: I switched it to use an f-string.

PR #400, opened 2024-04-17
Output bhr hangs for child processes

squarewave:
This PR just outputs an additional file containing hangs in child processes by piping through a config variable with the process name to filter on instead of hard-coding "Gecko."


squarewave:
( @BenWu )

BenWu (approved):
The changes make sense to me.  I started running it on my computer but it's taking a long time so it hasn't finished yet.  I'm happy to just merge it

PR #401, opened 2024-04-23
Parameterize bhr thread filtering

squarewave:
@BenWu 

I think we could also do with reducing the sampling - if my memory serves, there's a good chance this is still going to OOM without reducing it.

### DIFF
--- a/mozetl/bhr_collection/bhr_collection.py
+++ b/mozetl/bhr_collection/bhr_collection.py
@@ -1242,7 +1242,13 @@ def etl_job_daily(sc, sql_context, config=None):
     help="Proportion of pings to use (1.0 is 100%)",
 )
 @click.option("--use_gcs", is_flag=True, default=False)
-def start_job(date, sample_size, use_gcs):
+@click.option(
+    "--thread-filter", default="Gecko", help="Specifies which thread we are processing hangs from"
+)
+@click.option(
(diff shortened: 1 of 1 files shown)


BenWu (commented):
What would be an appropriate smaller sample size? We are just using the default 0.5 https://github.com/mozilla/telemetry-airflow/blob/main/dags/bhr_collection.py#L100

Can you run `black mozetl/bhr_collection/bhr_collection.py` to fix the CI?

mozetl/bhr_collection/bhr_collection.py
```diff
@@ -1242,7 +1242,13 @@ def etl_job_daily(sc, sql_context, config=None):
     help="Proportion of pings to use (1.0 is 100%)",
 )
 @click.option("--use_gcs", is_flag=True, default=False)
-def start_job(date, sample_size, use_gcs):
+@click.option(
+    "--thread-filter", default="Gecko", help="Specifies which thread we are processing hangs from"
+)
+@click.option(
+    "--output-tag", default="main", help="Tags the output filename"
+)
```
BenWu, line 1250: Format changes for ci
```suggestion
@click.option(
    "--thread-filter",
    default="Gecko",
    help="Specifies which thread we are processing hangs from",
)
@click.option("--output-tag", default="main", help="Tags the output filename")
```

BenWu (approved):

PR #403, opened 2024-09-04
bug(1914369): remove GRAPHICS_SANITY_TEST* related code from graphics dashboard due to related deprecation in gecko-dev

kik-kik:
# bug(1914369): remove GRAPHICS_SANITY_TEST* related code from graphics dashboard due to related deprecation in gecko-dev

The following commit in gecko-dev deprecated GRAPHICS_SANITY_TEST from `Histograms.json` which appears to have resulted in the ETL job using this python_mozetl logic to break as the result. We're expecting this code change to resolve this issue by the login no longer attempting to access the `GRAPHICS_SANITY_TEST` histogram.

### DIFF
--- a/mozetl/graphics/graphics_telemetry_dashboard.py
+++ b/mozetl/graphics/graphics_telemetry_dashboard.py
@@ -77,7 +77,6 @@ FeaturesKey = "environment/system/gfx/features"
 UserPrefsKey = "environment/settings/userPrefs"
 DeviceResetReasonKey = "payload/histograms/DEVICE_RESET_REASON"
 SANITY_TEST = "payload/histograms/GRAPHICS_SANITY_TEST"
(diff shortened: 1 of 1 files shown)


mozetl/graphics/graphics_telemetry_dashboard.py
```diff
@@ -96,8 +94,6 @@ def parse_args():
     MonitorsKey,
     ArchKey,
     DeviceResetReasonKey,
-    SANITY_TEST,
-    SANITY_TEST_REASON,
```
BenWu, line 100: We should keep `SANITY_TEST` since it doesn't expire (https://probes.telemetry.mozilla.org/?search=GRAPHICS_SANITY_TEST) and it's still used in the dashboard.  I think we should first try just removing things related to `SANITY_TEST_REASON` which is just the constants.

BenWu (changes requested):

BenWu (approved):
🤙

PR #404, opened 2024-09-10
Revert "bug(1914369): remove GRAPHICS_SANITY_TEST_REASON related code from graphics dashboard due to related deprecation in gecko-dev"

kik-kik:
As per comments in the bugzilla ticket, this should be added back and the change to gecko-dev removing this also needs to be reverted.


BenWu (approved):

BenWu (commented):
Actually reading the bug, `SANITY_TEST_REASON` might not get resolved soon so maybe we should leave it out for now and remove `PLUGIN_DRAWING_MODEL`, `D3D11_COMPOSITING_FAILURE_ID` and `OPENGL_COMPOSITING_FAILURE_ID` to get the job working again since we know those aren't needed

kik-kik:
@BenWu just noticed this still exists, do you think it makes sense to just close this considering it's been more than 6 months now with no action?

kik-kik:
Stale, closing. Will open a new PR if and when necessary.

PR #405, opened 2024-09-11
feat(bug-1917645): remove code related to PLUGIN_DRAWING_MODEL, D3D11_COMPOSITING_FAILURE_ID, and OPENGL_COMPOSITING_FAILURE_ID

kik-kik:
This is a follow-up to the following change in the Histograms.json in gecko-dev:

https://github.com/mozilla/gecko-dev/commit/7f25ee929ca325e057bdfc256bb978954b5fed0d#diff-6c55b27a94376254370c09aa7b397a80d4b2f7c3cb28b690f33557e92f9f98b8

This change removed the following from the list of histograms to process:

- `PLUGIN_DRAWING_MODEL`
- `D3D11_COMPOSITING_FAILURE_ID`
- `OPENGL_COMPOSITING_FAILURE_ID`

### DIFF
--- a/mozetl/graphics/graphics_telemetry_dashboard.py
+++ b/mozetl/graphics/graphics_telemetry_dashboard.py
@@ -80,10 +80,7 @@ SANITY_TEST = "payload/histograms/GRAPHICS_SANITY_TEST"
 STARTUP_TEST_KEY = "payload/histograms/GRAPHICS_DRIVER_STARTUP_TEST"
 WebGLSuccessKey = "payload/histograms/CANVAS_WEBGL_SUCCESS"
 WebGL2SuccessKey = "payload/histograms/CANVAS_WEBGL2_SUCCESS"
-PluginModelKey = "payload/histograms/PLUGIN_DRAWING_MODEL"
 MediaDecoderKey = "payload/histograms/MEDIA_DECODER_BACKEND_USED"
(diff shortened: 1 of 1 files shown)


mozetl/graphics/graphics_telemetry_dashboard.py
```diff
@@ -1061,7 +1042,6 @@ def advanced_layers_map(rdd):
         results = {
             "count": subset.count(),
             "compositors": subset.map(lambda p: (get_compositor(p),)).countByKey(),
-            "plugin_models": aggregate_plugin_models(subset),
```
BenWu, line 1064: This field is still used in the code that generates the graphs https://github.com/FirefoxGraphics/telemetry/blob/bb0cc53a61a30acb22b3973f32f1721c572d02bc/www/chart-graphs.js#L363

So we need to either remove it from the other repo or hardcode a value.  I would go with the quick solution of hardcoding a value since this isn't a high priority dashboard.

[The current json](https://storage.googleapis.com/moz-fx-data-static-websit-8565-analysis-output/gfx/telemetry-data/windows-features.json) has `[]` in one of the values and I think it just [renders nothing](https://firefoxgraphics.github.io/telemetry/#view=windows-features&winver=6.2.0) so an empty list should work

BenWu (approved):

Issue #409, opened 2026-04-10
Migrate bhr_collection.py to use glean data


squarewave:
We're already submitting a glean ping for the bhr data at `moz-fx-data-shared-prod.firefox_desktop_stable.hang_report_v1`. We should be able to modify `bhr_collection.py` to read from there without too much trouble.

PR #413, opened 2026-05-05
#409 Migrate BHR collection to Glean data

skylarkning:
## Summary

This PR refers to Issue #409.

This migrates `bhr_collection.py` to read BHR hang report data from the Glean table:

`moz-fx-data-shared-prod.firefox_desktop_stable.hang_report_v1`

instead of the legacy telemetry table:

`moz-fx-data-shared-prod.telemetry_stable.bhr_v4`.

## Changes

- Set `mozdata` as the BigQuery billing project  
- Read `client_info`, `ping_info`, and `metrics` from the Glean hang report table  
- Map Glean fields into the existing processing shape  
- Parse Glean hang report/module object metrics before processing  
- Support Glean stack frames shaped as `{frame, module}`  
- Remove the legacy `payload/time_since_last_ping` normalization path  

## Local Validation

Ran locally with:

```bash
JAVA_HOME=$(/usr/libexec/java_home -v 17) python3 ./mozetl/bhr_collection/bhr_collection.py \
  --date 2025-04-25 \
  --bq-connector-jar=spark-bigquery-latest_2.12.jar \
  --sample-size 0.0002
````

## Results and Comparison

The job completed successfully and wrote:

* `output/hangs_main_20250421.json`
* `output/hangs_main_current.json`

The output files are compared against legacy output:

* Top-level keys match
* Thread keys match
* `sampleTable` keys match
* Date keys match

The output is non-empty for the same thread/date combinations.

**Note:** Exact numeric similarity is no longer expected as the data source has changed to Glean, and the new output no longer normalizes by usage hours.

### DIFF
--- a/mozetl/bhr_collection/bhr_collection.py
+++ b/mozetl/bhr_collection/bhr_collection.py
@@ -551,7 +551,7 @@ def get_data(sc, sql_context, config, date, end_date=None):
         # sql results need to be saved to a table
         .option("materializationProject", "mozdata")
         .option("materializationDataset", "tmp")
-        .option("billingProject", "mozdata")
+        .option("billingProject", config["billing_project"])
         .load()
     )
 
@@ -1195,6 +1195,7 @@ default_config = {
     "exclude_modules": False,
     "uuid": uuid.uuid4().hex,
     "thread_filter": "Gecko",
+    "billing_project": "mozdata",
 }
 
 
@@ -1256,6 +1257,11 @@ def etl_job_daily(sc, sql_context, config=None):
 
 
 @click.command()
+@click.option(
+    "--billing-project",
+    default="mozdata",
+    help="GCP project to use for BigQuery billing.",
+)
 @click.option("--date", type=datetime.fromisoformat, required=True)
 @click.option(
     "--sample-size",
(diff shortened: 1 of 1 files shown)


squarewave:
@BenWu would you be able to take a look at reviewing this? Sky is a student worker who just started and he's going to be making some improvements to BHR, so we had him taking a look at migrating the bhr_collection job to use glean data as a first task. Regarding the coverage, I'd like to get some tests in but looking at what we have so far it looks like it might be a little involved to get a basic test of bhr_collection up and running. Does that sound right? In any case do you have any thoughts on the best route forward or salient examples for getting some bhr tests up and running either in this PR or in a follow-up?

BenWu (commented):
For testing, I think the simplest thing to do would be to mock the spark `load()` that gets data from bigquery with a dataframe with a few sample pings.  Then just run that through the rest of the job and verify the output.  That might be feasible but I'm not sure if spark complicates it.

Another note is that this job is currently using spark 2.4.8 via [dataproc image 1.5](https://github.com/mozilla/telemetry-airflow/blob/51f9defc13242c6d731038d4ed884c5569f09d3e/dags/bhr_collection.py#L72) which is [being discontinued in August this year](https://docs.cloud.google.com/dataproc/docs/concepts/versioning/dataproc-version-clusters#unsupported-image-versions) so this will need to be updated soon.  This is out of scope for this PR but I recommend doing a bit of research to get an idea of what would be required.  Bringing this up because I'm not sure we (on the data platform side) will be able to get to it before then.  That might be an easy AI job but the testing and validation is something I don't want to commit to right now

BenWu:
Github is having issues right now so I can't comment in the code

Can you make the billing project an input arg for the job? We'll still want to run in certain projects for the scheduled production jobs.

skylarkning:
> Github is having issues right now so I can't comment in the code
> 
> Can you make the billing project an input arg for the job? We'll still want to run in certain projects for the scheduled production jobs.

Hi @BenWu, thanks for the feedback and review! I added a `--billing-project` CLI option and pass it through the job config to the BigQuery connector. Default value is mozdata and scheduled production jobs can override it with a different project. 

Next I will look into the tests. Thank you!

BenWu (approved):
Just to confirm before I merge, is there anything else you want to add?

skylarkning:
> Just to confirm before I merge, is there anything else you want to add?

@BenWu Nothing in particular at this point. Thank you so much!

